PR #3, opened 2020-12-30
JCF: make sure git doesn't confuse users into thinking that auto-gene…

jcfreeman2:
…rated Nljs.hpp and Structs.hpp headers are meant to be manually edited and/or committed


bieryAtFnal:
I have confirmed that the addition of the .gitignore file avoids the confusion.

PR #5, opened 2021-01-05
Nested HDF5 groups in Utils

adam-abed-abud:
Hello! 

I pushed to the branch "NestedGroups" the SubGroup method in HDF5FileUtils. 
I also modified the write method of the HDF5DataStore and checked with the unit tests that everything was working fine.
These changes can also be added to the read method but I would prefer to wait and restore the read unittest. 

When implementing this feature I also thought one can also do something similar for datasets which would be useful for the TriggerRecordHeader. What do you think?

### DIFF
--- a/include/dfmodules/KeyedDataBlock.hpp
+++ b/include/dfmodules/KeyedDataBlock.hpp
@@ -29,6 +29,11 @@ public:
   const void* unowned_data_start;
   std::unique_ptr<char> owned_data_start;
 
+  // size_t trh_size;
+  //const void* unowned_trigger_record_header;
+  //std::unique_ptr<char> owned_trigger_record_header;
+
+
   explicit KeyedDataBlock(const StorageKey& theKey)
     : data_key(theKey)
   {}
@@ -42,7 +47,19 @@ public:
     }
   }
 
+/*
+  const void* getTriggerRecordHeader() const
+  {
+    if (owned_trigger_record_header.get() != nullptr) {
+      return static_cast<const void*>(owned_trigger_record_header.get());
+    } else {
+      return unowned_trigger_record_header;
+    }
+  }
+*/
+
   size_t getDataSizeBytes() const { return data_size; }
+  //size_t getTriggerRecordHeaderSizeBytes() const {return trh_size}
 };
 
 } // namespace dfmodules
--- a/plugins/DataGenerator.hpp
+++ b/plugins/DataGenerator.hpp
@@ -12,6 +12,7 @@
 #ifndef DFMODULES_SRC_DATAGENERATOR_HPP_
 #define DFMODULES_SRC_DATAGENERATOR_HPP_
 
+#include "dfmodules/CommonIssues.hpp"
 #include "dfmodules/DataStore.hpp"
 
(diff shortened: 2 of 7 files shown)


bieryAtFnal:
Hi Adam, which branch did you use as a starting point for the _NestedGroups_ branch?  It seems that there are many diffs on that branch compared to the _main_, _DataWriterModule_, and _feature/HDF5WriterConfig_ branches.

Maybe do you need to merge the _feature/HDF5WriterConfig_ branch into the _NestedGroups_ branch to pick up all of the recent changes?
Thanks,
Kurt

adam-abed-abud:
I started from the "feature/HDF5WriterConfig" branch and modified only two files (plugins/HDF5DataStore.hpp  and plugins/HDF5FileUtils.hpp). 

I will double check now.

Thanks
Adam

bieryAtFnal:
Yes, sorry, I realize what I did wrong.  I needed to update the other branches in my local copy of the repo.

bieryAtFnal:
The new function in HDF5FileUtils looks like a good start.  (For reference, in the protoDUNE HDF5 code, I created a function that could take the whole HDF5 "path".  That code is here:  https://cdcvs.fnal.gov/redmine/projects/dune-raw-data/repository/entry/dune-raw-data/HDFUtils/HDFFileHelper.cc?rev=feature%2Fsimple_hdf5_addition#L147 .  With this, the user only needs to make one call.)

I'm not sure what the question is regarding TriggerRecordHeader...
If we go the route of having StorageKey have special "none" values for APA number and link number when we store a TriggerRecordHeader, then the call to HDF5KeyTranslator::get_path_elements would only return two elements, and the function that I referenced above could be used to only create the necessary groups.

adam-abed-abud:
Hi Kurt,
Thanks a lot I liked the implementation you did in protoDUNE. I added it to the our HDF5 code.
I checked with the unit test and everything seems to be working fine. 

Regarding the dataset for the TriggerRecordHeader we could either do as you say or maybe a simpler way could be to define a method getTopGroup() and add the dataset from there. I will think about it a bit more before implementing anything. 

Thanks,
Adam

bieryAtFnal:
I see that there are currently methods in KeyedDataBlock related to the TriggerRecordHeader on this branch.  They are currently commented out.
KeyedDataBlock is not where we should be doing the special handling for the TRH. That should be in HDF5DataStore. 
Is that the plan?

bieryAtFnal:
Sorry, I'm not sure where the special handling for TriggerRecordHeader should go, but I would hope that we don't need to modify KeyedDataStore...

bieryAtFnal:
Adam and I talked about our ideas on how to handle TriggerRecordHeaders, and we have a plan for that.  
That topic is tangential to this pull request, though, so sorry for the noise.

For this PR, I have run a test with a slightly-real system with this code, and everything looks fine.
So, I'm going to approve this PR and merge the code to the main branch.

bieryAtFnal:
When I merged this PR, the merged code went into the feature/HDF5WriterConfig branch, so I merged the changes to the main branch by hand.

PR #7, opened 2021-01-07
Fragment receiver - first attempt

mroda88:


bieryAtFnal:
Hi Marco,
I tried running the code on this branch.  To do that, I first merged in the latest code on the main branch in this repo.
I used the latest code on the _main_ branch in the _dataformats_ repo in that test.  I needed to update a couple of minor things (e.g "APA" --> "apa").

Two things that I noticed:
* could we reduce the screen output?  Maybe make the burst of bookkeeping messages that we see at the beginning of the run TRACE message rather than ERS messages?
* only the first TriggerRecord is ever pushed to the output queue.  I added a TRACE statement right before the queue.push call, and only trigger number zero seems to get sent downstream.  This is why get get an Inhibit.

Do you want me to commit my local changes (e.g. the merge from main and the TRACE statement)?
Kurt

mroda88:
Hi Kurt, 
 the output is annoying, I know. We definitielly have to reduce it and lower the priority of the messages. Maybe for the first iteration we can leave it there. Or it will be a bit difficult to understand what is happening. 
Do you think it's my fault that only the first trigger record is pushed through? I usually see 6 before dying though. If you think it is may fault and it should need fixing before merging we can wait. If not please do the merge. 
Just be careful to use the latest commit: there should be a `while` loop around the push in my do_work().

mroda88:
Feel free to add all the local changes you need! I'll resume working tomorrow.

bieryAtFnal:
I made three categories of changes:
* I merged the recent changes on the _main_ branch into this branch
* I updated some "APA" names to match what is needed by the latest code in the dataformats repo ("apa")
* I modified FragmentReceiver.cpp to avoid a hang in the code after the first TriggerRecord is sent downstream.  Please see the comment in the code for a little more detail.

I have *not* merged this code to the _main_ branch because it depends on the latest _dataformats_ code, and I'm not sure that we want to convert to the new _dataformats_ until tomorrow.

As soon as we are ready for that, though, I think that this branch is ready.  Of course, it would be good to look a the change that I made to FragmentReceiver.cpp because that was a change to the logic.

PR #10, opened 2021-01-07
Feature/request generator

carloschavez:
PR contains the "real" RequestGenerator module and handles its basic configuration.
In order to run the minidaqapp the corresponding configuration in the minidaqapp schema should be used under feature/RequestGenerator branch in minidaqapp package.


bieryAtFnal:
I'm not sure how this was tested initially.
When I tried to run it, I got an Inhibit after a handful of events.
With some modifications to the slightlyreal application config, I was able to see data flow through the system and get written to disk.

carloschavez:
Hi Kurt, 
I was seeing an inhibit too after a few events , 
I was just testing that the requests were generated and sent to the fake readout app. 
What exactly did I missed in the configuration ?

PR #12, opened 2021-01-11
Feature/trigger record header

adam-abed-abud:
Hi all,

this branch contains the `AttemptCombo` from Marco as well as the required changes to be compatible with the "Trigger record header serialization" branch from `dataformtas`.

The TriggerRecordHeader has been added as a dataset to the DataWriter module and a unit test was created to test the functional aspect of theHDF5DataStore. 

Let me know if there are any changes I have to make. 

Thanks,
Adam


bieryAtFnal:
I tested the code on this branch using the set of branches from various repositories listed on [Phil's Integration Notes](https://github.com/DUNE-DAQ/minidaqapp/wiki/Phil-integration-notes-2021-01-07) (please see his notes for 11-Jan).

Things looked good with the HDF5/TriggerRecordHeader changes.  The only wrinkle that I noticed was the creation of many small files (e.g. one per fragment).  To work around this, I set a fairly large configuration value for max_file_size_bytes in minidaqapp/schema/minidaq-app-fake-readout.jsonnet.
I suspect that we should probably disable the max-file-size checking, for the moment, because it is currently willing/able to split individual events into different files, and we should have a discussion with offline and users before we make that the default behavior.

PR #16, opened 2021-01-18
Build against dunedaq-v2.1.0 release

alessandrothea:
Updates CMakeLists.txt and unittest's include paths to build against the `dunedaq-v2.1.0` release.


bieryAtFnal:
Both Adam and I have tested these changes, and they look good to us.

PR #23, opened 2021-02-02
Feature/issue20 hdf5dump changes

bieryAtFnal:
I believe that the changes on the feature/Issue20_hdf5dump_changes branch are at a useful stopping point and deserve to be merged (so that we can start making use of the initial human-readable conversion of the trigger_timestamp).

As noted in the associated Issue, the changes are in the hdf5_dump.py file, and they include changes from both Adam and me.  

To review this PR, I would suggest running the hdf5_dump.py script with both the -H and -TRH arguments.


eflumerf (approved):
Works well in -TRH and -H mode, with and without new -num argument

PR #26, opened 2021-02-04
Feature/refined stops

mroda88:
it implements the proper sequence of queue draining at the stop

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ find_package(HighFive REQUIRED)
 find_package(dfmessages REQUIRED)
 
 ##############################################################################
-daq_add_library( StorageKey.cpp TriggerDecisionForwarder TriggerInhibitAgent
+daq_add_library( StorageKey.cpp TriggerDecisionForwarder.cpp TriggerInhibitAgent.cpp
                  LINK_LIBRARIES 
                  ers::ers HighFive appfwk::appfwk stdc++fs dfmessages::dfmessages )
 
--- a/include/dfmodules/DataStore.hpp
+++ b/include/dfmodules/DataStore.hpp
@@ -16,12 +16,14 @@
 #define DFMODULES_INCLUDE_DFMODULES_DATASTORE_HPP_
 
 #include "dfmodules/KeyedDataBlock.hpp"
-#include "ers/ers.h"
-#include <appfwk/NamedObject.hpp>
-#include <nlohmann/json.hpp>
 
-#include <cetlib/BasicPluginFactory.h>
-#include <cetlib/compiler_macros.h>
+#include "appfwk/NamedObject.hpp"
+#include "cetlib/BasicPluginFactory.h"
+#include "cetlib/compiler_macros.h"
+#include "ers/ers.h"
+
+#include "nlohmann/json.hpp"
+
 #include <chrono>
 #include <cstddef>
 #include <memory>
@@ -89,7 +91,7 @@ public:
    * @brief Returns the list of all keys that currently existing in the DataStore
    * @return list of StorageKeys
    */
-  virtual std::vector<StorageKey> getAllExistingKeys() const = 0;
+  virtual std::vector<StorageKey> get_all_existing_keys() const = 0;
 
   // Ideas for future work...
   // virtual void write(const std::vector<KeyedDataBlock>& dataBlockList) = 0;
@@ -111,7 +113,7 @@ private:
  * @return unique_ptr to created DataStore instance
  */
 inline std::unique_ptr<DataStore>
-makeDataStore(const std::string& type, const nlohmann::json& conf)
+make_data_store(const std::string& type, const nlohmann::json& conf)
 {
   static cet::BasicPluginFactory bpf("duneDataStore", "make"); // NOLINT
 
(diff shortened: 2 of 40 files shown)


bieryAtFnal:
One quick question on the changes that I see in DataWriter.cpp...

Could the identical behavior be achieved with fewer modified lines of code, such as this?

```
--- a/plugins/DataWriter.cpp
+++ b/plugins/DataWriter.cpp
@@ -171,16 +171,19 @@ DataWriter::do_work(std::atomic<bool>& running_flag)
     throw InvalidDataWriterError(ERS_HERE, get_name());
   }
 
-  while (running_flag.load()) {
+  int number_of_empty_pop_calls = 0;
+  while (running_flag.load() || number_of_empty_pop_calls < 2) {
     std::unique_ptr<dataformats::TriggerRecord> trigger_record_ptr;
 
     // receive the next TriggerRecord
     try {
       m_trigger_record_input_queue->pop(trigger_record_ptr, m_queue_timeout);
       ++received_count;
+      number_of_empty_pop_calls = 0;
       TLOG(TLVL_WORK_STEPS) << get_name() << ": Popped the TriggerRecord for trigger number "
                             << trigger_record_ptr->get_header_ref().get_trigger_number() << " off the input queue";
     } catch (const dunedaq::appfwk::QueueTimeoutExpired& excpt) {
+      ++number_of_empty_pop_calls;
       // it is perfectly reasonable that there might be no data in the queue
       // some fraction of the times that we check, so we just continue on and try again
       continue;
```

bieryAtFnal:
Here are my corresponding proposed changes for RequestGenerator.cpp.  My sense is that do/while loops will always be advantageous in this code, even when not fully draining.

```
--- a/plugins/RequestGenerator.cpp
+++ b/plugins/RequestGenerator.cpp
@@ -134,21 +134,24 @@ RequestGenerator::do_work(std::atomic<bool>& running_flag)
     map[entry.first] = std::unique_ptr<datareqsink_t>(new datareqsink_t(entry.second));
   }
 
-  while (running_flag.load()) {
+  int number_of_empty_pop_calls = 0;
+  while (running_flag.load() || number_of_empty_pop_calls < 2) {  // or < 1
     dfmessages::TriggerDecision trigDecision;
     try {
       m_trigger_decision_input_queue->pop(trigDecision, m_queue_timeout);
       ++receivedCount;
+      number_of_empty_pop_calls = 0;
       TLOG(TLVL_WORK_STEPS) << get_name() << ": Popped the TriggerDecision for trigger number "
                             << trigDecision.m_trigger_number << " off the input queue";
     } catch (const dunedaq::appfwk::QueueTimeoutExpired& excpt) {
+      ++number_of_empty_pop_calls;
       // it is perfectly reasonable that there might be no data in the queue
       // some fraction of the times that we check, so we just continue on and try again
       continue;
     }
 
     bool wasSentSuccessfully = false;
-    while (!wasSentSuccessfully && running_flag.load()) {
+    do {
       TLOG(TLVL_WORK_STEPS) << get_name() << ": Pushing the TriggerDecision for trigger number "
                             << trigDecision.m_trigger_number << " onto the output queue";
       try {
@@ -164,6 +167,7 @@ RequestGenerator::do_work(std::atomic<bool>& running_flag)
           std::chrono::duration_cast<std::chrono::milliseconds>(m_queue_timeout).count()));
       }
     }
+    while (!wasSentSuccessfully && running_flag.load());
 
     //-----------------------------------------
     // Loop over trigger decision components
@@ -201,8 +205,7 @@ RequestGenerator::do_work(std::atomic<bool>& running_flag)
       auto& queue = it_req->second;
 
       wasSentSuccessfully = false;
-      while (!wasSentSuccessfully && running_flag.load()) {
-
+      do {
         TLOG(TLVL_WORK_STEPS) << get_name() << ": Pushing the DataRequest from trigger number "
                               << dataReq.m_trigger_number << " onto output queue :" << queue->get_name();
 
@@ -220,6 +223,7 @@ RequestGenerator::do_work(std::atomic<bool>& running_flag)
             std::chrono::duration_cast<std::chrono::milliseconds>(m_queue_timeout).count()));
         }
       }
+      while (!wasSentSuccessfully && running_flag.load());
     }
 
     m_trigger_decision_forwarder->set_latest_trigger_decision(trigDecision);
```

mroda88:
Hi @bieryAtFnal yes indeed it could. I considered it then I decided to separate explicitly the two loops, let's call them standard and drain loop respectively, because they might be different in the future. Let me try to explain. 
We discussed already that given the many graphs we can consider, there is no way to design a draining phase that does more ore less what the standard loop is doing. The best we can hope for is a clean way to simply clear the queues, without processing their content. As a consequence, I believe it is just accidental that the two operations are exactly identical in this case so a code that distinguishes the two phases might be beneficial in the future.
Another way to look at the same problem is with metrics or other log output. When the metrics are introduced, they will probably behave differently in the two phases. An example is the timing I added to the draining of the `FragmentReceiver`. We clearly don't care about how much time we stayed in the standard loop, but knowing the draining phase time will give us insight on the status of the queues until we have proper queue monitoring. 
That is why I specifically created different loops and wrapped common operations into functions to be used in the loops. 
I have more examples of non-similarities between the loops as part of my analysis of the possible crashes, I was planning to report at today's meeting.

bieryAtFnal:
Marco, thank you for making the code changes that we discussed in the MDApp/DF call today.
I tested those, and they worked as expected.
In addition, based on aspects of the discussion on the call, I simplified the loop test to use queue.can_pop in DataWriter and RequestGenerator.
I re-tested the system after those changes, and it still works as expected with a burst of 1000 triggers at stop time.  (I used Phil's trigemu::phililprodrigues/burst-at-stop branch for that, set the burst count in the config to 1000, and set the queue sizes to large values)

mroda88:
Seems good to me. Talk to you later.

bieryAtFnal (approved):
Several people have looked at these changes, we've discussed them within the DF sub-group within the MiniDAQ team, and they are in a good state.
In addition, both Marco and I (and maybe others) have run tests that confirm the correct behavior of the code at Stop time.

Issue #21, opened 2021-02-02
Some user-friendly additions to our data writing functionality


bieryAtFnal:
A couple of usability improvements in the data-writing part of the MiniDAQApp

1) Add the ability to disable the writing of data to disk

Giovanna has requested that this parameter be associated with the ‘start’ command.  That is, users should be able to enable and disable data-writing at the start of each run, without needing to reconfigure the system.

* I believe that we want this parameter to be associated with the DataWriter module (not the DataStore instance that is used by the DataWriter module).  In this way, *all* interactions with the DataStore instance can be disabled.  Given this, it might be better to name the parameter something like “data-storage-disable”.  (Not ‘disk-writing-disabled’.)

* A question that may come up:  should we even create a DataStore instance?  The answer to this must be “yes”, because the DataStore instance is created at ‘conf’ time, before we know whether data-storage will happen for a given run.

A bonus feature request from Giovanna:  for testing, it would also be cool to add a pre-scaling parameter, such that we may decide to write one event out of x. This is really only for test purposes and very long stability tests that we may want to do in the future, without lowering the rate of all the other data acquisition parts

2) Add unique identifiers to HDF5 filenames so that over-writes of existing files are prevented.

The unique identifiers that have been mentioned so far include a timestamp and a username.

This feature will naturally be part of the HDF5DataStore (not a general characteristics of DataStores, nor of the DataWriter) since it is focused on the names of the HDF5 data files.

We can imagine providing a configuration parameter for disabling this behavior, but it would be enabled by default.  

We’ll need to think about the right format for the timestamp.

3) Check for an invalid output directory or full disk at ‘Start’ time

This should be functionality within the HDF5DataStore.  However, it will need to be activated by a method call from the DataWriter module. It will likely require the addition of a new method in the DataStore interface, something like start_run().  Since some sort of ‘begin-run’ operation might be useful for other types of DataStores, it seems reasonable to add it to the interface.

We should throw an exception that somehow aborts the begin-run transition.

4) Close open files at Stop time

This should be functionality within the HDF5DataStore.  It will likely require the addition of a new method in the DataStore interface, something like stop_run().  Since some sort of ‘end-run’ operation might be useful for other types of DataStores, it seems reasonable to add it to the interface.

bieryAtFnal:
I have started making these changes on a branch in the _dfmodules_ repo.  That branch is based on the Style Guide change branch, so it already has those changes.

brettviren:
[...]

PR #28, opened 2021-02-05
Feature/issue21 data writing friendliness

bieryAtFnal:
I believe that the changes described in Issue #21 are done and are ready to be reviewed.  There are suggested tests listed in the comments of Issue #21.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ daq_add_library( StorageKey.cpp TriggerDecisionForwarder.cpp TriggerInhibitAgent
                  LINK_LIBRARIES 
                  ers::ers HighFive appfwk::appfwk stdc++fs dfmessages::dfmessages )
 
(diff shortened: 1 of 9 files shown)


carloschavez:
Hi Kurt,

I have ran the tests described in [issue 21](https://github.com/DUNE-DAQ/dfmodules/issues/21)
After your commit  
I have _checked_ that your fix for avoiding the prescale/Busy bug   is working. 

The other feature is still pending, when a very large max_file_size value is set (much > free space in the machine) and the error is not thrown at the start transition. 
But as I think this actually works, I am of the opinion that we could merge this into develop unless you prefer to fix this feature before merging, but I leave that up to you.

Thanks

bieryAtFnal (commented):
Carlos has run extensive tests to confirm that these changes work as described, and we've updated the code to fix the issues that he found.  As such, we believe that these changes are ready to go.

PR #31, opened 2021-02-10
Avoid crash in HDF5DataStore::write at beginning of second run in a MDApp instance

bieryAtFnal:
Giovanna and I both have seen crashes of the MDApp in the DataWriter when starting a second run in a single session.

I traced this to a missing test for an empty file ptr in HDF5DataStore::open_file_if_needed.  I should have found and fixed that problem when I added the prepare_for_run and finish_with_run methods, but I'm glad that we caught it now.

A candidate fix is on the bugfix/StopStartFilePtr branch, and the intention of this Pull Request is to ask someone to verify that the code change avoids the crash.

### DIFF
--- a/plugins/HDF5DataStore.hpp
+++ b/plugins/HDF5DataStore.hpp
@@ -349,7 +349,8 @@ private:
   void open_file_if_needed(const std::string& file_name, unsigned open_flags = HighFive::File::ReadOnly)
   {
 
-    if (m_basic_name_of_open_file.compare(file_name) || m_open_flags_of_open_file != open_flags) {
+    if (m_file_ptr.get() == nullptr || m_basic_name_of_open_file.compare(file_name) ||
+        m_open_flags_of_open_file != open_flags) {
 
       // 04-Feb-2021, KAB: adding unique substrings to the filename
       std::string unique_filename = file_name;


glehmannmiotto (changes requested):
This change avoids the crash, but after the second start the data writer complains that it cannot insert elements because they already exist, until it receives a trigger record number that had not been used in the previous run. So this is not the solution, yet.

bieryAtFnal:
I tried to reproduce the problem with existing elements, but I was unable to do so.  That is, I was able to successfully write TRs 1..5 in the first run, and 1..10 in the second run.  I tried running with both the stdin and rest cmdlibs.

It seems like somehow the file is not getting changed between runs in your environment.  (Also, you mentioned on Slack that you saw a crash from an existing filename.  That also gives a hint that there is an issue with generating a unique filename or opening a fresh file for each run.)  

We can trying debugging this together tomorrow.  Something that we can try is to enable the TRACE messages for HDF5DataStore and see what filenames it is creating at each Start transition.  (e.g. `dbt-setup-runtime-environment` ; `export TRACE_FILE=$DBT_INSTALL_DIR/../log/${USER}_dunedaq.trace` ; run the app ; `tshow | tdelta -ct 1 | grep HDF5D | grep prepare | more`)

philiprodrigues:
In the course of testing stop/start behaviour in trigemu, I pulled this branch. With `dfmodules` develop, I got crashes at the second "start"; with this branch those crashes are gone. I don't see any complaints about not being able to insert elements

bieryAtFnal (commented):
Several of us have successfully tested this fix, so I'm going to go ahead and merge it.

PR #32, opened 2021-02-11
Since dataformats now defines a TriggerRecordHeader_test, there is a

eflumerf:
target-name conflict with this test in dfmodules.


bieryAtFnal (approved):
I noticed the same problem.
Thanks for the fix.
I approve this change and will merge it into develop.

PR #35, opened 2021-02-12
Add python config for two-process application

philiprodrigues:
This PR adds a python script that produces the json configs needed to run a fake MiniDAQApp as two processes: one containing the TriggerDecisionEmulator and one containing everything else. The "fake" here means that there are no readout modules in the config: the fragments are provided by the FakeDataProd module from dfmodules.

There's also a python script to generate a config that's the same as `testapp_noreadout_confgen.py` but with pairs of `QueueToNetwork`/`NetworkToQueue` modules inserted in the connections with the `TriggerDecisionEmulator`. This is all still in one process, so its only purpose is to test the `QueueToNetwork`/`NetworkToQueue` functionality

### DIFF
--- a/python/dfmodules/testapp_noreadout_confgen.py
+++ b/python/dfmodules/testapp_noreadout_confgen.py
@@ -247,7 +247,7 @@ if __name__ == '__main__':
     @click.option('-d', '--data-file', type=click.Path(), default='./frames.bin')
     @click.option('-o', '--output-path', type=click.Path(), default='.')
     @click.option('--disable-data-storage', is_flag=True)
-    @click.argument('json_file', type=click.Path(), default='minidaq-app-fake-readout.json')
+    @click.argument('json_file', type=click.Path(), default='testapp_noreadout.json')
     def cli(number_of_data_producers, data_rate_slowdown_factor, run_number, trigger_rate_hz, data_file, output_path, disable_data_storage, json_file):
         """
           JSON_FILE: Input raw data file.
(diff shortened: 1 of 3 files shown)


bieryAtFnal:
I've compared the resulting data files from 

- testapp_noreadout_confgen.py, and 
- testapp_noreadout_networkqueue_confgen.py

and they are equivalent.  One request might be to reduce the verbosity of the ZmqSenderImpl  messages...

Also, testapp_noreadout_confgen.py has moved ahead of testapp_noreadout_networkqueue_confgen.py a bit (for example, with the ability to disable data writing).  Would it be worth adding those changes to testapp_noreadout_networkqueue_confgen.py?  I can help with that.

I'll look at testapp_noreadout_two_process.py next...

bieryAtFnal:
Same comments about testapp_noreadout_two_process.py: output file looks great, should we add disable_data_storage, etc.

philiprodrigues:
I agree that merging changes from `testapp_noreadout_confgen.py` into the other two configs is a good idea. If you have time to do that, that would be great (if you don't have time, I can add it to my todo list)

bieryAtFnal:
I will work on copying the recent additions from testapp_noreadout_confgen.py into the other two.

bieryAtFnal:
I believe that I have successfully added the --disable-data-storage command-line argument to testapp_noreadout_networkqueue_confgen.py and testapp_noreadout_two_process.py.

While I was making changes, I modified the default names of the resulting JSON files for all three python files.  If this change is undesirable, I can change that back.

philiprodrigues:
Thanks for the update. Works in my local tests.

Issue #33, opened 2021-02-11
dfmodulesConfig.cmake.in has a bug


jcfreeman2:
Alessandro noticed that `dfmodulesConfig.cmake.in` has the following line:
```
add_library(appfwk::appfwk ALIAS appfwk)
```
which suggests that `appfwkConfig.cmake` got copied into dfmodules and then got edited. What this means is that if you have a work area with a dfmodules repo and a repo which depends on dfmodules, you'll get an error. 

To fix this: the correct procedure to set up a package so it can be included as a dependency via `find_package(yourpackagename)` is described on the wiki, https://github.com/DUNE-DAQ/appfwk/wiki/Creating-a-new-package-under-v2.2.0, under "Installing your project as a local package".

bieryAtFnal:
I believe that this Issue can be closed thanks to Pull Request #37.

PR #37, opened 2021-02-15
Fix the Config.cmake.in

mroda88:
This *should*  be the correct fix for Issue #33


jcfreeman2:
Testing `find_package(dfmodules)` I discovered other problems not mentioned in Issue #33:
- `find_dependency(AppFwk)` was incorrect; replaced with `find_dependency(appfwk)`
- Since dfmessages is a dependency of dfmodules, I added `find_dependency(dfmessages)`

PR #39, opened 2021-02-17
Changes needed to send BufferTokens from DataWriter. Currently using the

eflumerf:
threshold_for_inhibit configuration parameter to determine how many
tokens to send at start.


eflumerf:
Note that this should be merged after `https://github.com/DUNE-DAQ/dfmessages/pull/14` and `https://github.com/DUNE-DAQ/trigemu/pull/15` (so that the testapp_noreadout configuration works correctly)

bieryAtFnal (approved):
looks good to me.
test with artificial delay in data writing limits rate, as expected.

Issue #24, opened 2021-02-02
A couple of minor changes to what we see in the output HDF5 file


bieryAtFnal:
At the moment, we are labeling the sub-detector Group in the HDF5 as "FELIX", when in fact, we should be labeling that "TPC".  It would be nice to change the HDF5DataWriter to output the "TPC" name.  There are other places where the "FELIX" name is also used, e.g. in unit tests.  Those are less important to change, but maybe we change those also, for completeness.

Another minor point is the APA number that we get when using the FakeDataProd module.  It seems like we get an APA number of 1 with it, whereas we get an APA number of zero when using the full v1.0 minidaqapp.  It would be nice to investigate if/why these are different and get the testapp_noreadout system to match the minidaq-app system.

Just a couple of minor things that I don't want to forget...

bieryAtFnal:
I made the changes for this Issue on branch feature/Issue24_MinorHDF5Changes.  That branch was based on the one for Issue34.  Once that branch is merged, I'll create a PR for this one.

glehmannmiotto:
I think it would be nice to introduce a lookup table with a clear association:
FragmentHeader. m_fragment_type -> Group name
This could then be used in the code and controlled from a single place.

bieryAtFnal:
I agree that it would be good to create a mapping from Fragment type to HDF5 Group name.  I will create an Issue in the minidaqapp repo for that, since it will affect multiple repositories.

In the meantime, I hope that we can get these simple changes approved and merged, so I will create a Pull Request in this repo for them.

PR #40, opened 2021-02-19
Feature/issue24 minor hdf5 changes

bieryAtFnal:
I believe that the changes for Issue #24 are ready to be reviewed.

### DIFF
--- a/plugins/DataGenerator.cpp
+++ b/plugins/DataGenerator.cpp
@@ -119,7 +119,7 @@ DataGenerator::do_work(std::atomic<bool>& running_flag)
   while (running_flag.load()) {
     for (size_t geo_id = 0; geo_id < m_geo_loc_count; ++geo_id) {
       // AAA: Component ID is fixed, to be changed later
-      StorageKey data_key(m_run_number, event_id, "FELIX", 0, geo_id);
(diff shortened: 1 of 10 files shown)


carloschavez:
Hi @bieryAtFnal Kurt,
I have check the changes and they all look fine to me.
I have tested and ran minidaq app successfully. 
I ran  grep on dfmodules for  "FELIX" and only found two harmless instances with that string in the following files :

schema/dfmodules-FakeDataProd-schema.jsonnet : doc="Temporary assignement of the FELIX link that is being read out (this will come from upstream soon)"),

test/disk_reader_demo.json:      "detector_id": "FELIX",

I guess we should either be consistent and chance the disk_reader_demo accordingly or remove the this configuration if it is not needed.
I can go ahead and merge this branch into develop as it is, or do you prefer to remove/change  disk_reader_demo.json ?

bieryAtFnal:
Hi @carloschavez , I made the change in disk_reader_demo.json.  Please check that it looks OK, and then merge, if you are OK with doing that.
Thanks,
Kurt

carloschavez:
Hi @bieryAtFnal, 
Thanks,   I will go ahead and merge this branch.
Carlos

Issue #30, opened 2021-02-10
Error Handling 


mroda88:
I'm opening this issue to keep track of the improvements we can implement for better error handling

### Data Writer
- [x] There is a `throw` in `do_work`, it should be substituted with a `fatal`. 
- [x] `do_init` and `do_conf` can throw. It is worth noting here that `make_data_store` can throw. 
- [x] the call to write https://github.com/DUNE-DAQ/dfmodules/blob/03e0dc38526734515846d9de86fd2e583dbbddae/plugins/DataWriter.cpp#L258    should be inside a try-catch loop as in principle other DataStores could throw. This call should also catch standard exceptions, not only ers.  

### HDF5DataStore
- [x] The catch block https://github.com/DUNE-DAQ/dfmodules/blob/03e0dc38526734515846d9de86fd2e583dbbddae/plugins/HDF5DataStore.hpp#L204-L208 can be simplified and made more consistent: we should only catch `HighFive` exception and rethrow something that can be cought by the DataWriter. 

### Fragment Receiever (clean)
* Here is the possibility that booking memory fails to be allocated if not enough resources are available in the machine but we agreed it is not worth guarding against this.

### KeyedDataBlock
- [x] following the discussion from yesterday, we could define the constructor as `noexcept`

glehmannmiotto:
About HDF5DataStore, I think that that library should take care of catching any std exceptions and creating/throwing ERS issues, allowing the higher levels of code to rely on the fact that catching ERS issues is sufficient.

philiprodrigues:
Re handling `std::bad_alloc`, I had thought that it was pointless because `bad_alloc` is only thrown when all is lost, but this article suggests that's not the case:

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1404r1.html

In particular, `bad_alloc` is likely to be thrown when requesting a stupendously large memory allocation, when smaller allocations (in the catch block, say) would still succeed.

I don't know what this implies for the specific case of handling `std::bad_alloc` in `FragmentReceiver`, but hopefully the article is useful in thinking it through.

mroda88:
Hi Phil, this is a nice input, thanks. I have two comments following the article you suggested. One is for the specific case of `std::bad_alloc` in our application, the other is related to the desired non-crashing behaviour desired from our applications. I think they are connected topics. 

[...]

PR #43, opened 2021-02-23
Desired error behaviour

mroda88:
possible solution for #30

### DIFF
--- a/cmake/dfmodulesConfig.cmake.in
+++ b/cmake/dfmodulesConfig.cmake.in
@@ -10,6 +10,7 @@ include(CMakeFindDependencyMacro)
 find_dependency(HighFive)
 find_dependency(appfwk)
 find_dependency(dfmessages)
+find_dependency(logging)
 
 # Figure out whether or not this dependency is an installed package or
 # in repo form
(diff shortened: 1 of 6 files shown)


mroda88:
Note that I added a missing addition from Ron in cmake/dfmodulesConfig.cmake.in

bieryAtFnal:
Marco, thanks for adding that entry in cmake/dfmodulesConfig.cmake.in!

bieryAtFnal:
The changes look good to me.  I made a couple of additional changes, please check them:  I ran clang-format, and I added a try/catch around open_file_if_needed calls in HDF5DataStore.hpp.

The reason for the latter:  as part of my testing, I specified a bogus output path ("/abc").  With this, HDF5DataStore and DataWriter correctly threw an exception at Start time that complained that it was a bogus path.  That was all good.
To test how one type of write() error might be handled, I then commented out that check of the output path at Start time.  With that, I noticed an uncaught std::exception from the HighFive code when it tried to open a new file at the bogus path during a write() call.  To protect against this, I added the try/catch.

mroda88:
Good catches! I'll have a look before the meeting today

bieryAtFnal (approved):
These changes look good to me.

PR #45, opened 2021-02-24
Remove m_ prefix from struct public members. Change window_offset/window_width to window_begin/window_end

eflumerf:


bieryAtFnal (approved):
LGTM

PR #46, opened 2021-02-25
Use daq-codegen from updated daq-cmake and demonstrate use of opmonlib

bieryAtFnal:
These code changes add the daq-codegen call to the CMakeLists.txt file, move and rename the schema/*.jsonnet files, and add opmonlib calls to DataWriter.


eflumerf (approved):
LGTM

PR #47, opened 2021-03-02
Moved data_storage_prescale to Conf transition; cleaned up old-style jsonnet and json files

bieryAtFnal:
A big part of the changes are the removal of obsolete files...


eflumerf:
The code builds and the python modules generate json files. However, I'm not sure what the intended use-case is for data_generator_demo and data_combiner_demo, so while I was able to run them through the daq_application state machine, I don't know if they did what they were supposed to do or how to check that.

eflumerf (approved):

bieryAtFnal:
Thanks, Eric.
Those data_generator and data_combiner are carry-overs from our distributed-event-building tests in ddpdemo.  I want to keep them around, and I hope to restore their functionality at this point, but they are probably not yet fully updated, so it's good that you didn't spend too much time on them.
I'm going to merge this code to _develop_ now.  There will be forthcoming changes, though, to get testapp_noreadout_networkqueue and testapp_noreadout_two_process to work.

PR #48, opened 2021-03-03
Reducing the CPU usage of the FragmentReceiver

mroda88:
this should reduce the CPU consumption.
At the moment we cannot compile due to missing dependencies, hence the pull request is just a draft. Once we'll be able to compile I'll make the pull request a full one.


mroda88:
Compiles and works.

bieryAtFnal (approved):
The change looks good to me, and I also ran the test of looking at the CPU from the FragmentReceiver (ffr) thread before and after the change and confirming that it is no longer high after the change.

PR #52, opened 2021-03-05
Made use of fragment_type_to_string function; removed obsolete detector_name config param

bieryAtFnal:
I ran a test in which I buggered the fragment ID that is put into each Fragment in the readout code and noticed that we weren't yet making full use of the fragment_type _to_string functionality.  So, I added that to the DataWriter module, and removed the obsolete detector_name configuration parameter from the hdf5datastore.jsonnet schema file.

I tested the changes with buggered fragment types in the readout code (tested kPDS and value==22).  I saw the expected error in the latter case.


mroda88 (approved):
For me it's good to go!

PR #53, opened 2021-03-08
Changed window offset and width to window begin and end in hdf5dump.py

bieryAtFnal:


scripts/hdf5dump/hdf5_dump.py
```diff
@@ -91,8 +91,10 @@ def print_frag_header(data_array):
   print("Trig number:\t\t", bytes_to_int((trigger_number)[::-1]))
   print("Trig timestamp:\t\t", bytes_to_int((trigger_timestamp)[::-1]),
         "("+str(datetime.datetime.fromtimestamp(float(bytes_to_int((trigger_timestamp)[::-1]))/50000000.0))+")")
-  print("Window offset:\t\t", bytes_to_int((window_offset)[::-1]))
-  print("Window width:\t\t", bytes_to_int((window_width)[::-1]))
+  print("Window begin:\t\t", bytes_to_int((window_begin)[::-1]),
+        "("+str(datetime.datetime.fromtimestamp(float(bytes_to_int((window_begin)[::-1]))/50000000.0))+")")
+  print("Window end:\t\t", bytes_to_int((window_end)[::-1]),
+        "("+str(datetime.datetime.fromtimestamp(float(bytes_to_int((window_end)[::-1]))/50000000.0))+")")
```
eflumerf, line 97: This constant here...is this something that the configuration python might alter? Would it be a good idea to make it an option/declare it somewhere so it can be changed?

bieryAtFnal:
Yes, we should make it a constant that is assigned at the top of the file.  Beyond that, I expect that it may need to change if/when the clock frequency changes.  There has been some discussion about how to indicate that in the data, but that discussion didn't reach a conclusion, as far as I know, so a constant at the top of the script may be fine, for now.

mroda88 (approved):
I was about to write exactly what Eric wrote. I think it's good to go

PR #54, opened 2021-03-09
Make two-process readout work with processes on different hosts

philiprodrigues:
This PR makes the two-process fake readout work with the processes on different hosts. The IP addresses of the trigemu and DF hosts are passed via the `--host-ip-trigemu` and `--host-ip-df` command-line arguments respectively


bieryAtFnal (approved):
These changes look good to me, and I have run successful tests on the mu2edaq cluster at Fermilab.  In those tests, I accepted the default host-ip values, I specified a single host for both, and I specified different hosts for each.  In all cases, TriggerDecisions and fake time sync messages flowed between the two processes, as evidenced by the expected data being written to disk.

PR #56, opened 2021-03-11
Fix for bug in which small max_file_size_bytes values were splitting TriggerRecords

bieryAtFnal:
When I decreased the value of the max_file_size_bytes parameter below the size of an individual TriggerRecord, but larger than a single Fragment, I saw TriggerRecord Fragments get written into separate files.  This violates what I've told the Offline to expect.

I believe that the code changes on this branch will fix the problem.  They may not be the final ones related to writing data blocks and opening files, though, since we need to get back to some more general 'data store'  functionality at some point.  That will not be in dunedaq-v2.4.0, though.

### DIFF
--- a/plugins/HDF5DataStore.hpp
+++ b/plugins/HDF5DataStore.hpp
@@ -201,10 +201,7 @@ public:
   virtual void write(const KeyedDataBlock& data_block)
   {
     // check if a new file should be opened for this data block
-    if ((m_recorded_size + data_block.m_data_size) > m_max_file_size) {
-      ++m_file_index;
-      m_recorded_size = 0;
-    }
+    increment_file_index_if_needed(data_block.m_data_size);
 
     // determine the filename from Storage Key + configuration parameters
     std::string full_filename = HDF5KeyTranslator::get_file_name(data_block.m_data_key, m_config_params, m_file_index);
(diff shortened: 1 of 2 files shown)


bieryAtFnal:
I realized that I should add some unit test conditions for small file size limits, so I will do that now.

mroda88:
Hi @bieryAtFnal , I think the changes are ok. In the sense that they are doing what we desired. But I think we are unprotected against a too small size, which is hard to judge. Since the size of a datablock is runtime dependent, we might end up in the situation in which we create an empty file, then we updated the index and we write the data in a new file while having created an empty file for nothing. This is annoying because it might create a lot of empty files, and I don't see the offline being happy about this either. 
I think the condition for updating the file index should be updated from
```C++
   if ((m_recorded_size + data_block.m_data_size) > m_max_file_size) {
      ++m_file_index;
      m_recorded_size = 0;
    }
```
To something like
```C++
   if ((m_recorded_size + data_block.m_data_size) > m_max_file_size && m_recorded_size > 0 ) {
      ++m_file_index;
      m_recorded_size = 0;
    }
```
Clearly this violate the maximum size requirement. So Alternatively we should throw an error that states that the max size file prevents us to write a specific trigger record.
What do you think?
If we go in the first direction, I would suggest to write a protected function that deals with this conditions as it is used in 2 different places. 
I hope I managed to explain myself, otherwise pop a message here or on slack.

bieryAtFnal:
Hi Marco,
That seems like a good safety check.  I've committed changes based on your suggestions.
The two unit tests that focus on file-size-checking still pass, so that is good.  I'll run some 'system' tests too.

mroda88 (approved):
good to go for me

PR #58, opened 2021-03-17
Fragment receiver metrics

mroda88:
Following yesterday's request from Giovanna, here are the desired metrics. Already tested and working.

### DIFF
--- a/plugins/FragmentReceiver.cpp
+++ b/plugins/FragmentReceiver.cpp
@@ -106,7 +106,7 @@ FragmentReceiver::init(const data_t& init_data)
 }
 
 void 
-get_info(opmonlib::InfoCollector& ci, /*int level*/)  {
+FragmentReceiver::get_info(opmonlib::InfoCollector& ci, int /*level*/)  {
 
   fragmentreceiverinfo::Info i;
 
@@ -116,14 +116,6 @@ get_info(opmonlib::InfoCollector& ci, /*int level*/)  {
 
   ci.add(i) ;
 
-}
-
-records_received = m_records_received_tot.load();
-  dwi.new_records_received = m_records_received.exchange(0);
-  dwi.records_written = m_records_written_tot.load();
-  dwi.new_records_written = m_records_written.exchange(0);
-
-  ci.add(dwi);
 }
 
 void
@@ -187,7 +179,7 @@ FragmentReceiver::do_work(std::atomic<bool>& running_flag)
     m_trigger_decisions_counter.store( m_trigger_decisions.size() ) ;
     m_fragment_index_counter.store( m_fragments.size() ) ;
     uint64_t tot = std::accumulate( m_fragments.begin(), m_fragments.end(), 0, 
-				    [&](auto tot, auto ele){ return tot += ele.size() ; } ) ;
+				    [&](auto tot, auto & ele){ return tot += ele.second.size() ; } ) ;
     m_fragment_counter.store( tot ) ;
     
 
--- a/plugins/FragmentReceiver.cpp
+++ b/plugins/FragmentReceiver.cpp
@@ -12,6 +12,7 @@
 #include "appfwk/DAQModuleHelper.hpp"
 #include "appfwk/app/Nljs.hpp"
 #include "dfmodules/fragmentreceiver/Nljs.hpp"
+#include "dfmodules/fragmentreceiverinfo/Nljs.hpp"
 #include "dfmodules/fragmentreceiver/Structs.hpp"
 #include "logging/Logging.hpp"
 
@@ -22,6 +23,7 @@
 #include <thread>
 #include <utility>
 #include <vector>
+#include <numeric>
 
 /**
  * @brief TRACE debug levels used in this source file
@@ -103,6 +105,27 @@ FragmentReceiver::init(const data_t& init_data)
   TLOG_DEBUG(TLVL_ENTER_EXIT_METHODS) << get_name() << ": Exiting init() method";
 }
 
+void 
+get_info(opmonlib::InfoCollector& ci, /*int level*/)  {
+
+  fragmentreceiverinfo::Info i;
+
+  i.trigger_decisions = m_trigger_decisions_counter.load() ; 
+  i.trigger_fragments = m_fragment_index_counter.load() ;
+  i.total_fragments = m_fragment_counter.load() ;
+
+  ci.add(i) ;
+
+}
+
+records_received = m_records_received_tot.load();
+  dwi.new_records_received = m_records_received.exchange(0);
+  dwi.records_written = m_records_written_tot.load();
+  dwi.new_records_written = m_records_written.exchange(0);
+
+  ci.add(dwi);
+}
+
 void
 FragmentReceiver::do_conf(const data_t& payload)
 {
@@ -145,6 +168,10 @@ FragmentReceiver::do_work(std::atomic<bool>& running_flag)
   m_trigger_decisions.clear();
   m_fragments.clear();
 
+  m_trigger_decisions_counter.store( 0 ) ;
+  m_fragment_index_counter.store( 0 ) ;
+  m_fragment_counter.store( 0 ) ;
+
   // allocate queues
   trigger_decision_source_t decision_source(m_trigger_decision_source_name);
   trigger_record_sink_t record_sink(m_trigger_record_sink_name);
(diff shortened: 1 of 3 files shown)


bieryAtFnal (approved):
The changes look good to me.  I ran a test with 10 links and a slowdown factor of 1 on a computer that couldn't quite keep up with the necessary amount of work, and I saw non-zero values for the new statistics during the run, and these stats went back to zero after the run ended.  And, a different run that did not stress the computer showed zero values for these stats during the whole run.

bieryAtFnal:
Hi Marco,
In one of my tests after the addition of the "old_fragments" metric, I saw the following output from the periodic onmon printout:

```
    "/mdapp/ffr/fragmentreceiverinfo/data/class_name": "fragmentreceiverinfo",
    "/mdapp/ffr/fragmentreceiverinfo/data/old_fragments": 9,
    "/mdapp/ffr/fragmentreceiverinfo/data/total_fragments": 63,
    "/mdapp/ffr/fragmentreceiverinfo/data/trigger_decisions": 7,
    "/mdapp/ffr/fragmentreceiverinfo/data/trigger_fragments": 7,
```

At first, I was confused by the fact that the old_fragment count was larger than the trigger_decision count, but then I slowed down and realized that the first quantity is old *fragments*.  So, the right comparison is 9 old_fragments vs. 63 total_fragments, right?

A question that I have is whether "old_trigger_decisions" or "old_trigger_fragments" might be more intuitive than "old_fragments".  I realize that this is based on my expectation that we would be interested in the number of stale TriggerRecords, and others may have different expectations.  Thoughts?

mroda88:
So, to the first question the answer is yes: you are very correct. 

About the naming. In principle we can even have both, "old_trigger_decisions" and "old_trigger_fragments". But we need to decide what is old for a trigger decision. So I suspect we can leave "old_trigger_decisions" aside for now. I'm happy to rename "old_fragments" to "old_trigger_fragments" if it makes things easier. I'm not sure it will change the order in which they appear as I suspect it's alphabetical. Shall I go on?

mroda88:
But then, if we rename "old_trigger_fragments" can it be that people will compare that to "trigger_fragments" ? In that case we should also rename "trigger fragments" to "Fragment indexes" or something like that

bieryAtFnal:
If I understand correctly, "old_trigger_fragments" would have a different calculation associated with it from "old_fragments".  So, we shouldn't change the name if we don't change the calculation.  
But, yes, "old_trigger_fragments" is probably what I was thinking about when I first started my testing of the latest changes.  That is, the number of TriggerRecords (i.e. TriggerDecisions plus one or more Fragments) in the book that have one or more Fragments that are too old.

mroda88:
Ah, I see. That might be a useful information as well. Shall I added?

bieryAtFnal:
Yes, please.

mroda88:
It would look like this
```bash
mdapp_test/ffr/fragmentreceiverinfo/data/old_fragments": 235,
    "/mdapp_test/ffr/fragmentreceiverinfo/data/old_trigger_ids": 8,
    "/mdapp_test/ffr/fragmentreceiverinfo/data/populated_trigger_ids": 10,
    "/mdapp_test/ffr/fragmentreceiverinfo/data/total_fragments": 313,
    "/mdapp_test/ffr/fragmentreceiverinfo/data/trigger_decisions": 10,

```

mroda88:
I pushed the commit. If we like it, I can update the wiki page once it's merged.

bieryAtFnal:
During a run, the additional metric showed what I expected.  Here is a sample:

```
    "/mdapp/ffr/fragmentreceiverinfo/data/class_name": "fragmentreceiverinfo",
    "/mdapp/ffr/fragmentreceiverinfo/data/old_fragments": 62,
    "/mdapp/ffr/fragmentreceiverinfo/data/old_trigger_ids": 7,
    "/mdapp/ffr/fragmentreceiverinfo/data/populated_trigger_ids": 9,
    "/mdapp/ffr/fragmentreceiverinfo/data/total_fragments": 71,
    "/mdapp/ffr/fragmentreceiverinfo/data/trigger_decisions": 9,
```

A question, though.  After I ended the run by using the stop command without a pause first, I see the following:

```
    "/mdapp/datawriter/datawriterinfo/data/class_name": "datawriterinfo",
    "/mdapp/datawriter/datawriterinfo/data/new_records_received": 0,
    "/mdapp/datawriter/datawriterinfo/data/new_records_written": 0,
    "/mdapp/datawriter/datawriterinfo/data/records_received": 145,
    "/mdapp/datawriter/datawriterinfo/data/records_written": 145,
    "/mdapp/datawriter/datawriterinfo/time": 1616086406,
    "/mdapp/ffr/fragmentreceiverinfo/data/class_name": "fragmentreceiverinfo",
    "/mdapp/ffr/fragmentreceiverinfo/data/old_fragments": 40,
    "/mdapp/ffr/fragmentreceiverinfo/data/old_trigger_ids": 6,
    "/mdapp/ffr/fragmentreceiverinfo/data/populated_trigger_ids": 0,
    "/mdapp/ffr/fragmentreceiverinfo/data/total_fragments": 0,
    "/mdapp/ffr/fragmentreceiverinfo/data/trigger_decisions": 0,
    "/mdapp/ffr/fragmentreceiverinfo/time": 1616086406,
    "/mdapp/tde/triggerdecisionemulatorinfo/data/class_name": "triggerdecisionemulatorinfo",
    "/mdapp/tde/triggerdecisionemulatorinfo/data/inhibited": 79,
    "/mdapp/tde/triggerdecisionemulatorinfo/data/new_inhibited": 0,
    "/mdapp/tde/triggerdecisionemulatorinfo/data/new_triggers": 0,
    "/mdapp/tde/triggerdecisionemulatorinfo/data/triggers": 145,
    "/mdapp/tde/triggerdecisionemulatorinfo/time": 1616086406
```

The fact that triggerdecisionemulatorinfo/data/triggers and datawriterinfo/data/records_written matched was encouraging.

When I looked at the data file using our hdf5dump.py tools, I noticed that there were 9 TRs with an error bit set, and those 9 TRs had one or more missing Fragments.  In total, I believe that there were 25 missing fragments.  

Is there an straightforward way to correlate the 9 TRs with errors and 25 missing fragments in the data file with the following onmon values?
```
    "/mdapp/ffr/fragmentreceiverinfo/data/old_fragments": 40,
    "/mdapp/ffr/fragmentreceiverinfo/data/old_trigger_ids": 6,
```
Thanks!

mroda88:
That's a very good question. I think there is but not without introducing new metrics. 
The missing information (for the user) is how many fragments should be collected for every trigger decision. This is an information we have inside the fragment receiver, in fact it is the information we use to decide when a trigger record is complete. I wonder what useful metric we could implement here. 

I have a few ideas, but I'm not convinced by any of them.
* missing_fragments : this can be exactly the count you are asking, but this metric will be always 0, except at the end of run. That's why it seems more a matter of error reporting to me, rather than a continuous stream of info. And it might happen that the metric interrogator does not even have the time to request this number before it is useful.
* expected_fragments :  this can be the number of fragments yet to be received in total with respect to those requested. More useful as it changes with time but then when the trigger records are forcefully created at the `stop` it goes to zero because the trigger decisions disappears. This is also an indirect measure of the quantity you are asking, not enough to predict exactly what you mentioned, but it might give an order of magnitude.

I would prefer the second. What I suspect will happen is the following. In normal conditions you should have 0 or a low number as the trigger_decisions are down to zero. Under stress, trigger_decisions and expected_fragments will grow. In case of missing fragments expected fragments will rise, when you call the stop the number will collapse, but having a graphic tool you can see the cliff and have an idea of the missing fragments.

What do you think?

bieryAtFnal:
I didn't think that I was asking for new metrics. but after looking at the code and thinking about what you wrote, I now understand why you're suggesting them.  As you've indicated, "too old" doesn't necessarily correlate to "eventually incomplete".  For example, at the end of the run there could be a TR that ends up incomplete, but its timestamp is quite recent.

Thinking as I write...  As we've said, there isn't a direct connection between 

- _old_fragments_ and _old_trigger_ids_ at the end of the run
- and
- _the number of missing_fragments_ and _the number of TRs with missing fragments_ in the files on disk

but there is a reasonable chance that the following will be true:

- the number of TRs on disk with missing fragments >= number of old_trigger_ids at the end of the run
- number of missing fragments in the data on disk >= ((old_trigger_ids * expected_fragments_per_TR) - old_fragments)

But, none of that is guaranteed, so it's mainly just me writing some notes to help me think it through.

For the purposes of this PR, I vote that we keep the metrics that you've added as they are, and we think about any 'incomplete TR' metrics separately

mroda88:
Yes, those correlations are not guaranteed at all. And yes I agree with what you propose. Following what we said at the meeting I opened an issue with two proposed new metrics for DataWriter to be used for the missing fragment purpose. Let's expand that list with what we think we'll need.

PR #60, opened 2021-03-24
Several unit tests were updated to make use of configuration structs

bieryAtFnal:
I've updated several unit tests that are in current use to make use of our configuration structs when creating HDF5DataStore instances. The goal is to gain the usual advantages of the configuration structs, namely compile-time checking of the struct data members and default values for the data members that do not need to be directly specified.

I tried to make these sorts of changes in earlier versions of the code, but the underlying tools (e.g. daq-cmake) didn't have the right features yet.  But, now they do, so these changes can go in.

These changes will also need to be made in HDF5Combiner_test.cxx and HDF5GetAllKeys_test.cxx, but since those tests are not in current use (they need more basic changes needed first), I deferred making these changes in those files until later.

### DIFF
--- a/unittest/HDF5Read_test.cxx
+++ b/unittest/HDF5Read_test.cxx
@@ -7,7 +7,9 @@
  * received with this code.
  */
 
-#include "HDF5DataStore.hpp"
+#include "dfmodules/DataStore.hpp"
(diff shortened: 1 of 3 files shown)


mroda88 (approved):
good for me

PR #65, opened 2021-04-10
Dingpf/issue 55 hdf5dump feature

dingp:
The new helper looks like:

```sh
bash $ ./hdf5_dump.py -h
usage: hdf5_dump.py [-h] -f FILE_NAME [--header {trigger,fragment,both}] [--check-fragments] [--list-components] [-n NUM_OF_RECORDS] [-v]

Python script to parse DUNE-DAQ HDF5 output files.

optional arguments:
  -h, --help            show this help message and exit
  -f FILE_NAME, --file_name FILE_NAME
                        Path to HDF5 file
  --header {trigger,fragment,both}
                        Select which header data to display
  --check-fragments     check if fragments written in trigger record matches expected number in trigger record header
  --list-components     list components in trigger record header
  -n NUM_OF_RECORDS, --num-of-records NUM_OF_RECORDS
                        Select number of records to be parsed
  -v, --version         show program's version number and exit
```

The new option `--list-components`, when used together with `--header trigger`, will list the expected components in the trigger record header:

```sh
bash $ ./hdf5_dump.py --header trigger -n 1 -f sample.hdf5 --list-components
Reading file sample.hdf5
================================================================================
Path                          : TriggerRecord00000/TriggerRecordHeader
Size                          : (288, 1)
Data type                     : int8
Magic word                    : 0x33334444
Version                       : 1
Trigger number                : 0
Trigger timestamp             : 79759095071000000 (2020-07-19 13:05:01.420000)
No. of requested components   : 10
Run Number                    : 333
Error bits                    : 0
Trigger type                  : 0
--------------------------------------------------------------------------------
GeoID (APA)                   : 0
GeoID (link)                  : 0
Begin time (ticks@50MHz)      : 1000
End time (ticks@50MHz)        : 1200
--------------------------------------------------------------------------------
GeoID (APA)                   : 0
GeoID (link)                  : 1
Begin time (ticks@50MHz)      : 1000
End time (ticks@50MHz)        : 1200
--------------------------------------------------------------------------------
GeoID (APA)                   : 0
GeoID (link)                  : 2
Begin time (ticks@50MHz)      : 1000
End time (ticks@50MHz)        : 1200
--------------------------------------------------------------------------------
GeoID (APA)                   : 0
GeoID (link)                  : 3
Begin time (ticks@50MHz)      : 1000
End time (ticks@50MHz)        : 1200
--------------------------------------------------------------------------------
GeoID (APA)                   : 0
GeoID (link)                  : 4
Begin time (ticks@50MHz)      : 1000
End time (ticks@50MHz)        : 1200
--------------------------------------------------------------------------------
GeoID (APA)                   : 0
GeoID (link)                  : 5
Begin time (ticks@50MHz)      : 1000
End time (ticks@50MHz)        : 1200
--------------------------------------------------------------------------------
GeoID (APA)                   : 0
GeoID (link)                  : 6
Begin time (ticks@50MHz)      : 1000
End time (ticks@50MHz)        : 1200
--------------------------------------------------------------------------------
GeoID (APA)                   : 0
GeoID (link)                  : 7
Begin time (ticks@50MHz)      : 1000
End time (ticks@50MHz)        : 1200
--------------------------------------------------------------------------------
GeoID (APA)                   : 0
GeoID (link)                  : 8
Begin time (ticks@50MHz)      : 1000
End time (ticks@50MHz)        : 1200
--------------------------------------------------------------------------------
GeoID (APA)                   : 0
GeoID (link)                  : 9
Begin time (ticks@50MHz)      : 1000
End time (ticks@50MHz)        : 1200
```

The new option `--check-fragments` will go through the whole file and list the expected number of components obtained from the trigger record header, as well as those actually written into the corresponding trigger record.

```sh
bash $ ./hdf5_dump.py --check-fragments -f sample.hdf5
Reading file sample.hdf5
---------------------Column Definitions---------------------
i:           Trigger record number;
N_frag_exp:  expected no. of fragments stored in header;
N_frag_act:  no. of fragments written in trigger record;
N_diff:      N_frag_act - N_frag_exp
---------------------Column Definitions---------------------
    i       N_frag_exp     N_frag_act     N_diff
    0           10             10           0
    1           10             10           0
    2           10             10           0
    3           10             10           0
    4           10             10           0
    5           10             10           0
    6           10             10           0
    7           10             10           0
    8           10             10           0
    9           10             10           0
    10          10             10           0
    11          10             10           0
    12          10             10           0
    13          10             10           0
    14          10             10           0
    15          10             10           0
    16          10             10           0
    17          10             10           0
    18          10             10           0
    19          10             10           0
    20          10             10           0
    21          10             10           0
    22          10             10           0
    23          10             10           0
    24          10             10           0
    25          10             10           0
    26          10             10           0
    27          10             10           0
```

### DIFF
--- a/scripts/hdf5dump/hdf5_dump.py
+++ b/scripts/hdf5dump/hdf5_dump.py
@@ -13,37 +13,6 @@ import datetime
 
 CLOCK_SPEED_HZ = 50000000.0
 
-def parse_args():
-    parser = argparse.ArgumentParser(description='Python script to parse DUNE-DAQ HDF5 output files.')
-    parser.version = '0.1'
-    parser.add_argument('-p', '--path',
-                       help='Path to HDF5 file',
-                       default='./')
-
-    parser.add_argument('-f', '--file_name',
-                       help='Name of HDF5 file',
-                       required=True)
-
-    parser.add_argument('-H', '--header_only',
-                       help='Print the header only; no data is displayed',
-                       action='store_true')
-
-    parser.add_argument('-TRH', '--TRH_only',
-                       help='Print the TriggerRecordheader only; data is displayed',
-                       action='store_true')
-
-    parser.add_argument('--check-fragment',
-                       help='check if fragments written in trigger record matches expected number in trigger record header',
-                       action='store_true')
-
-    parser.add_argument('-num', '--num_req_trig_records',
-                        type=int,
-                       help='Select number of trigger records to be parsed',
-                       default=None)
-
-    parser.add_argument('-v', '--version', action='version')
-    return parser.parse_args()
-
 ## Print FragmentHeader
 def print_frag_header(data_array):
   (check_word, version, fragment_size,
@@ -88,14 +57,15 @@ def print_trh(data_array):
   print("Error bits:\t\t", error_bits)
   print("Trigger type:\t\t", trigger_type)
 
-  j = 0
-  for (i_geo_id_apa, i_geo_id_link, i_begin, i_end) in struct.iter_unpack("<2I2Q", data_array[48:]):
-      print("Expected Component: \t\t", j)
-      print("       GeoID (APA): \t\t", i_geo_id_apa)
-      print("       GeoID (link): \t\t", i_geo_id_link)
-      print("       Begin time (ticks@50MHz): \t", i_begin)
-      print("       End time (ticks@50MHz): \t\t", i_end)
-      j += 1
+  if listComponents:
+      j = 0
+      for (i_geo_id_apa, i_geo_id_link, i_begin, i_end) in struct.iter_unpack("<2I2Q", data_array[48:]):
+          print("Expected Component: \t\t", j)
+          print("       GeoID (APA): \t\t", i_geo_id_apa)
+          print("       GeoID (link): \t\t", i_geo_id_link)
+          #print("       Begin time (ticks@50MHz): \t", i_begin)
+          #print("       End time (ticks@50MHz): \t\t", i_end)
+          j += 1
 
   return
 
@@ -115,7 +85,6 @@ def examine_fragments(name, dset):
         trigger_record_fragments[nheader] += 1
     if isinstance(dset, h5py.Dataset) and "TriggerRecordHeader" in name:
         # This is a new TriggerRecordHeader
-        print("New TriggerRecordHeader : ", name)
         data_array = bytearray(dset[:])
         (i,) = struct.unpack('<Q', data_array[24:32])
         trigger_record_header_expected_fragments.append(i)
@@ -126,7 +95,7 @@ def print_header_func(name, dset):
     if nheader < NHEADER and isinstance(dset, h5py.Dataset):
         if printTRH and "TriggerRecordHeader" in name:
             print(80*'=')
-            print("nheader:\t\t", nheader)
+            print("TriggerRecordHeader:\t\t", nheader)
             print('Path:\t\t', name)
             print('Size:\t\t', dset.shape)
             print('Data type:\t\t', dset.dtype)
@@ -135,7 +104,7 @@ def print_header_func(name, dset):
             print_trh(data_array)
         if not printTRH and "TriggerRecordHeader" not in name:
             print(80*'=')
-            print("nheader:\t\t", nheader)
+            print("Fragment:\t\t", nheader)
             print('Path:\t\t', name)
             print('Size:\t\t', dset.shape)
             print('Data type:\t\t', dset.dtype)
@@ -144,9 +113,6 @@ def print_header_func(name, dset):
             print_frag_header(data_array)
     return
 
-nheader = 0
-NHEADER = 0
-printTRH = False
 def get_header(file_name):
     with h5py.File(file_name, 'r') as f:
         f.visititems(print_header_func)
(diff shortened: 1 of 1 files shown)


dingp:
Commits up to ca551f38cd78014c38b66adafa3ef2dc6a00f606 (those on April 9th) preserved code of original script as much as possible.

Commits after those are mostly refactoring the script, and change the coding style.

We haven't set up a style guide for python though. I was taking this chance to make the script PEP8 compliant and getting a feeling of it. Once we agree on the style, a CI workflow can even be set up to check the style compliance.

adam-abed-abud:
Hi Pengfei,

I tested the new changes and from a functional point of view all of them work fine. I have just a few minor comments: 

"python3 hdf5_dump.py --header  both  -f sample.hdf5  -n 1": this prints only the fragment and not "fragment+trigger" for only 1 record

"python3 hdf5_dump.py  -f swtest_run41340849_0000_biery_20210413T134930.hdf5 --list-components -n 5":  this prints only 1 trigger record instead of 5. It only works by adding "--header trigger" as you wrote in the comment. I think it is better to specify this in the helper or better fix it in the code if it is possible

"python3 hdf5_dump.py --header  both  -f swtest_run41340849_0000_biery_2021043T134930.hdf5   --check-fragments": this is not parsing properly the actual number of fragments

I tested with the sample.hdf5 file which is in the repository and with another hdf5 file which you can find here if you want to reproduce the issue: /afs/cern.ch/user/a/aabedabu/work_public/check_parsing_script/swtest_run41340849_0000_biery_20210413T134930.hdf5

Adam

dingp:
@adam-abed-abud  Thanks a lot for the careful check!


> "python3 hdf5_dump.py --header both -f sample.hdf5 -n 1": this prints only the fragment and not "fragment+trigger" for only 1 record
> 
> "python3 hdf5_dump.py -f swtest_run41340849_0000_biery_20210413T134930.hdf5 --list-components -n 5": this prints only 1 trigger record instead of 5. It only works by adding "--header trigger" as you wrote in the comment. I think it is better to specify this in the helper or better fix it in the code if it is possible

The number of records option here is a bit confusing. When `--header both` is used, it actually counts against both the fragment headers and trigger headers. So in `-n 1` is used with `--header both`, only the first record which is the fragment header is printed.

In the second test, the output actually contains one trigger header and four fragment header since `--header both` is the default value.

I can remove the default `both` and set `--header` as a required argument, and add a bit more info in the helper for `-n` to emphasize that it is the total number of records for either `TriggerRecordHeader` or `FragmentHeader`, depending on which option is used with `--header`.

> 
> "python3 hdf5_dump.py --header both -f swtest_run41340849_0000_biery_2021043T134930.hdf5 --check-fragments": this is not parsing properly the actual number of fragments
> 
> I tested with the sample.hdf5 file which is in the repository and with another hdf5 file which you can find here if you want to reproduce the issue: /afs/cern.ch/user/a/aabedabu/work_public/check_parsing_script/swtest_run41340849_0000_biery_20210413T134930.hdf5
> 

Thanks for finding this! I was parsing `sample.hdf5` only when developing the script, and `FELIX` is always used in the path to fragments. I thought that was all. But the file you tested with did not have `FELIX` in the path, instead, it uses `TPC`. I've removed the filter on path name with keywords as `FELIX` or `TPC`.

Another note, `--check-components` ignores options of `--header` as well as `-n`. When that turns on, the check goes through the whole file. I can either add this info to the helper, or actually make it work with `-n`.


> Adam

bieryAtFnal:
I don't know if it is too late to be asking this question, but could the -n parameter be designed so that it always specifies the number of TriggerRecords?
If this were true, then we might lose the functionality to only look at N of M total fragments in the first TriggerRecord, but that doesn't seem like a big loss to me.
It seems like we would benefit from -n always meaning the same thing, and not have it depend on choice of other parameters.

dingp:
> I don't know if it is too late to be asking this question, but could the -n parameter be designed so that it always specifies the number of TriggerRecords?

@bieryAtFnal Indeed. I realized the same. I've updated my comments above while implementing the fixes. 

`-n` is now counting the TriggerRecords, unless `--header=fragment` is specified. It is a step closer. Let me know if you prefer to count against trigger records only, even if `--header=fragment` is used. 

I felt like it is not necessary since one can use `--header=both -n 2` to print both trigger records and fragments if they really just want to see all the fragments of two trigger records, and the additional trigger record headers in the output are only a small fraction of all the output.

dingp:
@bieryAtFnal  In the latest commit, I made `-n` count against trigger records, and updated the helper message to say it is for `Select number of trigger records to be parsed`. I agree that this is the better way to go since it is clearer.

dingp:
Here is a possible improvement of the script's options for our discussion tomorrow: change `--header` to `--print-header`, and remove `none` as one of its choices. So users will know that it is explicitly used for printing header info.

The script's helper will look like the following. It will additionally require at least one of `--print-header` and `check-components` is used.

```sh
$ ./hdf5_dump.py -h
usage: hdf5_dump.py [-h] -f FILE_NAME [-p {trigger,fragment,both}] [-c] [-l] [-n NUM_OF_RECORDS] [-v]

Python script to parse DUNE-DAQ HDF5 output files.

optional arguments:
  -h, --help            show this help message and exit
  -f FILE_NAME, --file_name FILE_NAME
                        Path to HDF5 file
  -p {trigger,fragment,both}, --print-header {trigger,fragment,both}
                        Select which header data to display
  -c, --check-fragments
                        check if fragments written in trigger record matches expected number in trigger record header
  -l, --list-components
                        list components in trigger record header, used together with "--print-header trigger" or "--print-header both"
  -n NUM_OF_RECORDS, --num-of-records NUM_OF_RECORDS
                        Select number of trigger records to be parsed
  -v, --version         show program's version number and exit

$ ./hdf5_dump.py -f sample.hdf5
Error: use at least one of the two following options:
       -p, --print-header {trigger,fragment,both}
       -c, --check-fragments
```

dingp:
Proposed interface changed implemented in the last commit 96f87b6.

Looking into how we can print `TriggerRecrodHeader` before `FragmentHeader`, I feel like the easiest way (without buffering the output and reorganizing before printing out), is to:
1. use `h5py.visit` to traverse through the paths in the file, and save all paths of datasets for the number of trigger records requested;
2. the paths will be saved into a list like: `dump_paths = [{"TriggerRecordHeader": "TriggerRecordHeader0001, "Fragments": ["List of Fragment_Paths"]) ], in which the element type is a dictionary. 
3. loop over the saved paths, and call the print function to process the headers according to the option used by `--print-header`.

dingp:
I was replying on `sample.hdf5` when developing the script, and forgot it might be obsolete. For example, the fragment begin/end time in `sample.hdf5` is `1000` and `1200` respectively. It is the duration counted as no. of ticks instead of actual timestamp.

In a more recent data file, e.g. `swtest_run41340849_0000_biery_20210413T134930.hdf5`, those fields are proper timestamp. I've updated the script to do the timestamp conversion for those fields now.

bieryAtFnal (approved):
The changes look great to me.

adam-abed-abud:
Successfully tested the latest changes. Going to merge this branch.

PR #67, opened 2021-04-27
TRBuilder

mroda88:


bieryAtFnal:
Marco, I re-merged the `develop` branch in _dfmodules_ into the `mroda/TRBuilder` branch.  This only affected documentation, a CI yml file, and the hdf5_dump.py script.  So, it should not have affected your changes at all.

mroda88:
I pulled the new code and recompiled. No issues.

bieryAtFnal:
I've run tests with artificial delays in the DataWriter module (to simulate some sort of disk-writing slowdown), and the results of those tests looked good.
Some info:

- the artificial delay was implemented as a usleep(1370000) statement in DataWriter.cpp near the DataStore::write() call (the configured trigger rate was 1 Hz, so the 1.37 second delay created a backlog of TRs)
- I ran tests with two settings of the token count, 10 and 23.
- In all cases, I saw the expected number_of_elements in the trigger_record_q (approximately equal to the token count minus one)
- in all cases, the expected number of TriggerRecords were successfully written to the file (demonstrating successful draining of the system)

I did see some cases in which the nanorc "stop" command timed out, but I believe that this was simply the result of the timeout value being set too low for a situation in which it took 34 seconds to drain the events out of the system.

mroda88:
Ok, I removed the Fake plugins and Request Generator and FragmentReceiver. I was not sure what to do with DataGenerator, but I noticed it was used by the python script so I left it there. 
If we are happy with this, we are probably ready to merge and we can work on the fix necessary to use the new GeoID.

mroda88:
Question, do we want to add the feature added in dfmessages with https://github.com/DUNE-DAQ/dfmessages/pull/27/ ?

bieryAtFnal:
I like the idea of handling the change from request_mode to readout type in a separate dfmodules PR.

I will run a test with the latest code on the mroda/TRBuilder branch, and if it looks good, we can merge it to develop in the next hour.  Is that OK?

mroda88:
Perfect!

bieryAtFnal:
My test looked fine.

bieryAtFnal (approved):
Looks good to me

PR #70, opened 2021-05-05
Update GeoID usages

eflumerf:
This PR is for compatibility with DUNE-DAQ/dataformats#62 (which should be merged first)

### DIFF
--- a/plugins/DataWriter.cpp
+++ b/plugins/DataWriter.cpp
@@ -234,7 +234,7 @@ DataWriter::do_work(std::atomic<bool>& running_flag)
 
         // print out some debug information, if requested
         TLOG_DEBUG(TLVL_FRAGMENT_HEADER_DUMP)
-          << get_name() << ": Partial(?) contents of the Fragment from link " << frag_ptr->get_link_id().link_number;
+          << get_name() << ": Partial(?) contents of the Fragment from link " << frag_ptr->get_link_id().element_id;
         const size_t number_of_32bit_values_per_row = 5;
         const size_t max_number_of_rows = 5;
         int number_of_32bit_values_to_print =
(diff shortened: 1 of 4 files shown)


plugins/DataWriter.cpp
```diff
@@ -234,7 +234,7 @@ DataWriter::do_work(std::atomic<bool>& running_flag)
 
         // print out some debug information, if requested
         TLOG_DEBUG(TLVL_FRAGMENT_HEADER_DUMP)
-          << get_name() << ": Partial(?) contents of the Fragment from link " << frag_ptr->get_link_id().link_number;
+          << get_name() << ": Partial(?) contents of the Fragment from link " << frag_ptr->get_link_id().element_id;
```
mroda88, line 237: I think here we can simply do
```C++
frag_ptr->get_link_id()
```
Now that we have the operator <<

plugins/DataWriter.cpp
```diff
@@ -264,8 +264,8 @@ DataWriter::do_work(std::atomic<bool>& running_flag)
         StorageKey fragment_skey(frag_ptr->get_run_number(),
                                  frag_ptr->get_trigger_number(),
                                  dataformats::fragment_type_to_string(frag_ptr->get_fragment_type()),
-                                 frag_ptr->get_link_id().apa_number,
-                                 frag_ptr->get_link_id().link_number);
+                                 frag_ptr->get_link_id().region_id,
```
mroda88, line 267: Here we can updated the ERS issue accepting directly the GedID

plugins/RequestGenerator.cpp
```diff
@@ -88,8 +88,9 @@ RequestGenerator::do_conf(const data_t& payload)
 
   for (auto const& entry : parsed_conf.map) {
     dataformats::GeoID key;
-    key.apa_number = entry.apa;
-    key.link_number = entry.link;
+    key.component_type = dataformats::GeoIDComponentType::kTPC;
```
mroda88, line 91: This will need major changes in the schema as well to add the type, I think. But since the whole code is deprecated I would leave it as it is to not break the compilation.

mroda88 (commented):
Just added a few comments, we don't have to act on them now

mroda88:
I would wait for these changes until we merge the TRBuilder. then we fix things, I'll be happy to take care of it if necessary: the changes go beyond renaming and there is a bit of logic to be coded.

PR #74, opened 2021-05-07
Proper GeoID implementation

mroda88:
I could not compile all the repos, so this is just the beginning. 
But this version should have the proper changes in the schema as well, to be compatible with https://github.com/DUNE-DAQ/dataformats/pull/62


mroda88:
This is a bit crude for the DataWriter part.

eflumerf (approved):
LGTM

PR #76, opened 2021-05-17
Rename Fragment::get_link_id and set_link_id to get_element_id and set_element_id

eflumerf:
Depends on DUNE-DAQ/dataformats#66


bieryAtFnal (approved):
LGTM

PR #77, opened 2021-05-20
Support for PDS and other system types in the HDF5 file

bieryAtFnal:
This PR is for some initial changes to support PDS and other subdetectors (_system types_) in the HDF5 raw data files that we write.

The code changes are still somewhat rough.  I have plans for better ways of doing things, and this is just the first step (to get something working this week).  I'll want to describe the better plans to the MiniDAQ Dev group sometime soon.

In the meantime, I believe that these changes provide the necessary functionality.  I've made some assumptions that are probably not ideal, and we can update them as we learn more.  For example, for PDS data, I set the HDF5 Group equal to "PDS", region-based Groups to "Region000", and DataSets to "Element00".   This is to be compared to the TPC, which has "TPC", "APA000", and "Link00".

To test these changes, I made some hacks in other repositories to create some mis-labelled PDS data.  And, I only used a two-link system in my tests.

Here is the _readout_ hack:
```
diff --git a/src/WIBRequestHandler.hpp b/src/WIBRequestHandler.hpp
index 752a2ff..cc8be83 100644
--- a/src/WIBRequestHandler.hpp
+++ b/src/WIBRequestHandler.hpp
@@ -70,8 +70,13 @@ protected:
     fh.window_begin = dr.window_begin;
     fh.window_end = dr.window_end;
     fh.run_number = dr.run_number;
+    if ((m_link_number % 2) == 0) {
       fh.element_id = { dataformats::GeoID::SystemType::kTPC, m_apa_number, m_link_number };
       fh.fragment_type = static_cast<dataformats::fragment_type_t>(dataformats::FragmentType::kTPCData);
+    } else {
+      fh.element_id = { dataformats::GeoID::SystemType::kPDS, m_apa_number, m_link_number };
+      fh.fragment_type = static_cast<dataformats::fragment_type_t>(dataformats::FragmentType::kPDSData);
+    }
     return std::move(fh);
   } 
```

Here is the _trigger_ hack:
```
diff --git a/.github/workflows/dunedaq-develop-cpp-ci.yml b/.github/workflows/dunedaq-develop-cpp-ci.yml
(dbt-pyvenv) [biery@mu2edaq13 sourcecode]$ cd trigger
(dbt-pyvenv) [biery@mu2edaq13 trigger]$ git diff
diff --git a/plugins/ModuleLevelTrigger.cpp b/plugins/ModuleLevelTrigger.cpp
index 76ab008..91e88df 100644
--- a/plugins/ModuleLevelTrigger.cpp
+++ b/plugins/ModuleLevelTrigger.cpp
@@ -76,9 +76,15 @@ ModuleLevelTrigger::do_configure(const nlohmann::json& confobj)
   m_initial_tokens = params.initial_token_count;
 
   m_links.clear();
+  int idx = 0;
   for (auto const& link : params.links) {
     // For the future: Set APA properly
-    m_links.push_back(dfmessages::GeoID{dataformats::GeoID::SystemType::kTPC, 0, static_cast<uint32_t>(link) }); // NOLINT
+    if ((idx % 2) == 0) {
+      m_links.push_back(dfmessages::GeoID{dataformats::GeoID::SystemType::kTPC, 0, static_cast<uint32_t>(link) }); // NOLINT
+    } else {
+      m_links.push_back(dfmessages::GeoID{dataformats::GeoID::SystemType::kPDS, 0, static_cast<uint32_t>(link) }); // NOLINT
+    }
+    ++idx;
   }
 
   m_configured_flag.store(true);
```

Here is the _minidaqapp_ hack:
```
diff --git a/python/minidaqapp/nanorc/dataflow_gen.py b/python/minidaqapp/nanorc/dataflow_gen.py
index b58b449..dcb7f3a 100644
--- a/python/minidaqapp/nanorc/dataflow_gen.py
+++ b/python/minidaqapp/nanorc/dataflow_gen.py
@@ -136,8 +136,11 @@ def generate(NETWORK_ENDPOINTS,
         
                 ("trb", trb.ConfParams( general_queue_timeout=QUEUE_POP_WAIT_MS,
                                         map=trb.mapgeoidqueue([
-                                                trb.geoidinst(region=0, element=idx, system="TPC", queueinstance=f"data_requests_{idx}")  for
+                                                trb.geoidinst(region=0, element=0, system="TPC", queueinstance=f"data_requests_0"),
+                                                trb.geoidinst(region=0, element=1, system="PDS", queueinstance=f"data_requests_1") ]
+                                            #for idx in range(NUMBER_OF_DATA_PRODUCERS) ]
                                                               ))),
+
                 ("datawriter", dw.ConfParams(initial_token_count=TOKEN_COUNT,
                             data_store_parameters=hdf5ds.ConfParams(name="data_store",
                                 # type = "HDF5DataStore", # default
```
(Remember that you need to re-generate your system configuration [e.g. _mdapp_4proc_] after you make this confgen change *and* rebuild.)

Some additional notes on the code changes on this branch:

- the number of bytes written by the DataWriter was added to its opmon stats
- the run number was added to the Start and Stop TRACE messages in the DataWriter
- two HDF5 Attributes have been added to the raw data files.  These are intended to help us decode the HDF5 formatting of the raw data as it evolves.
   - data_format_version (set to 1)
   - operational_environment (set to "swtest")
- some obsolete code was removed

Another item that is still on the to-do list is to update the unit tests (some of these have been commented out, for expediency).


include/dfmodules/StorageKey.hpp
```diff
@@ -1,66 +1,69 @@
...
-  inline static const std::string s_invalid_detector_type = "Invalid";
-  static constexpr int s_invalid_apa_number = std::numeric_limits<int>::max();
-  static constexpr int s_invalid_link_number = std::numeric_limits<int>::max();
+  static constexpr int s_invalid_region_number = std::numeric_limits<int>::max();
+  static constexpr int s_invalid_element_number = std::numeric_limits<int>::max();
+
+  /**
+   * @brief The group that should be used within the data record.
+   */
+  enum DataRecordGroupType
```
mroda88, line 32: Will this enum scare offiline people?

eflumerf (approved):
Code changes look good. Builds and runs in default configuration as expected.

mroda88 (commented):
Hi Kurt, this is enlightening, at least. I didn't realise we had to introduce specifications related to iceberg or the environment. So I didn't (and I don't) have ideas or opinions here: I need to think about it and read carefully what the code is doing.
I absolutely agree that a discussion is necessary to have people on the same page.

As you were saying, as a start, this look good to me. 
I'm making notes for the discussions here. "Notes" means open question to me, you don't have to answer :-) 
 * What are the things you think are going to change in the future? Is it up to us to make a proposal or do we need feedback from at least offline?
 * I wonder, if there are ways to reduce the amount of code we have to write every time we introduce a subsystem or update the GeoID. Could we have a system where a key is simply a string + a TR ID and the problem of creating the string is for other pieces of code? Like the TR or the fragments themselves? Would this model be more sustainable? 
* How much this will change for the near detector systems? I don't think they will have only one or 2 subsystems there, given the diversity of subdetectors. How does this code scale with that? If it scales.

mroda88 (approved):
As I was writing in the comment the code looks ok to me, so I'll add an approve

bieryAtFnal:
Hi Marco,
As you say, the "operational environment" is something that we should talk about.  I'm not totally sure that it will be needed, so it's inclusion at this point is more of a precaution.  I can imagine slightly different formatting rules in different environments, though (for example, number of digits in the runNumber field in the filename could have N digits in one environment and M digits in another).

The StorageKey class is not something seen by offline, so they shouldn't have an opinion, IMO.

Issue #78, opened 2021-05-20
DUNE Linter Cleanup


eflumerf:
Note: I ran `dbt-build.sh --lint` on the `kbiery/PdsSupportInHDF5` branch (PR #77)

```
sourcecode/dfmodules/include/dfmodules/DataStore.hpp:35:  #define appears to be used. Macros should generally be avoided if there's an alternative to them.  [build/define_used] [3]
sourcecode/dfmodules/include/dfmodules/DataStore.hpp:43:  #define appears to be used. Macros should generally be avoided if there's an alternative to them.  [build/define_used] [3]
sourcecode/dfmodules/include/dfmodules/StorageKey.hpp:21:7: warning: class 'StorageKey' defines a non-default destructor but does not define a copy constructor, a copy assignment operator, a move constructor or a move assignment operator [cppcoreguidelines-special-member-functions]
sourcecode/dfmodules/include/dfmodules/StorageKey.hpp:21:7: warning: class 'StorageKey' defines a non-default destructor but does not define a copy constructor, a copy assignment operator, a move constructor or a move assignment operator [cppcoreguidelines-special-member-functions]
sourcecode/dfmodules/include/dfmodules/StorageKey.hpp:21:7: warning: class 'StorageKey' defines a non-default destructor but does not define a copy constructor, a copy assignment operator, a move constructor or a move assignment operator [cppcoreguidelines-special-member-functions]
sourcecode/dfmodules/include/dfmodules/StorageKey.hpp:21:7: warning: class 'StorageKey' defines a non-default destructor but does not define a copy constructor, a copy assignment operator, a move constructor or a move assignment operator [cppcoreguidelines-special-member-functions]
sourcecode/dfmodules/plugins/DataWriter.cpp:217:  An unsigned integer appears to be used here.  [build/unsigned] [3]
sourcecode/dfmodules/plugins/DataWriter.cpp:301:  Small and focused functions are preferred: DataWriter::do_work() has 93 non-comment lines (error triggered by exceeding 80 lines) (disable this by putting "// NOLINT" after the function's closing brace.)  [readability/fn_size] [1]
sourcecode/dfmodules/plugins/DataWriter.hpp:77:  An unsigned integer appears to be used here.  [build/unsigned] [3]
sourcecode/dfmodules/plugins/DataWriter.hpp:78:  An unsigned integer appears to be used here.  [build/unsigned] [3]
[...]

PR #80, opened 2021-05-20
Run clang-format on files, implement some lint fixes.

eflumerf:
This fixes the trivial items in #78, however, a few more substantive complaints remain.

```
sourcecode/dfmodules/plugins/DataWriter.cpp:301:  Small and focused functions are preferred: DataWriter::do_work() has 93 non-comment lines (error triggered by exceeding 80 lines) (disable this by putting "// NOLINT" after the function's closing brace.)  [readability/fn_size] [1]
sourcecode/dfmodules/plugins/HDF5DataStore.hpp:166:  A catch-all-exceptions construct was found in a file which doesn't contain int main(); this can only be used to wrap main()  [runtime/exceptions] [5]
sourcecode/dfmodules/plugins/HDF5DataStore.hpp:214:  A catch-all-exceptions construct was found in a file which doesn't contain int main(); this can only be used to wrap main()  [runtime/exceptions] [5]
sourcecode/dfmodules/plugins/HDF5DataStore.hpp:252:  A catch-all-exceptions construct was found in a file which doesn't contain int main(); this can only be used to wrap main()  [runtime/exceptions] [5]
sourcecode/dfmodules/plugins/HDF5DataStore.hpp:431:  A catch-all-exceptions construct was found in a file which doesn't contain int main(); this can only be used to wrap main()  [runtime/exceptions] [5]
sourcecode/dfmodules/plugins/TriggerRecordBuilder.cpp:343:  Small and focused functions are preferred: TriggerRecordBuilder::do_work() has 88 non-comment lines (error triggered by exceeding 80 lines) (disable this by putting "// NOLINT" after the function's closing brace.)  [readability/fn_size] [1]
```

### DIFF
--- a/include/dfmodules/DataStore.hpp
+++ b/include/dfmodules/DataStore.hpp
@@ -32,6 +32,7 @@
 #include <vector>
 
 #ifndef EXTERN_C_FUNC_DECLARE_START
+// NOLINTNEXTLINE(build/define_used)
 #define EXTERN_C_FUNC_DECLARE_START                                                                                    \
   extern "C"                                                                                                           \
   {
@@ -40,6 +41,7 @@
  * @brief Declare the function that will be called by the plugin loader
  * @param klass Class to be defined as a DUNE IPM Receiver
  */
+// NOLINTNEXTLINE(build/define_used)
 #define DEFINE_DUNE_DATA_STORE(klass)                                                                                  \
   EXTERN_C_FUNC_DECLARE_START                                                                                          \
   std::unique_ptr<dunedaq::dfmodules::DataStore> make(const nlohmann::json& conf)                                      \
--- a/include/dfmodules/StorageKey.hpp
+++ b/include/dfmodules/StorageKey.hpp
@@ -50,7 +50,6 @@ public:
     , m_element_number(element_number)
 
   {}
-  ~StorageKey() {} // NOLINT
 
   int get_run_number() const;
   int get_trigger_number() const;
--- a/plugins/DataWriter.cpp
+++ b/plugins/DataWriter.cpp
@@ -298,7 +298,7 @@ DataWriter::do_work(std::atomic<bool>& running_flag)
   }
 
   TLOG_DEBUG(TLVL_ENTER_EXIT_METHODS) << get_name() << ": Exiting do_work() method";
-}
+} // NOLINT(readability/fn_size)
 
 } // namespace dfmodules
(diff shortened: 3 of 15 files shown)


jcfreeman2:
The existing changes (don't rely on indirect `#include`ing of files, etc.) look good. As far as the `catch(...)` construct: in the Styleguide's [section on exceptions](https://dune-daq-sw.readthedocs.io/en/latest/packages/styleguide/#73-exceptions), the following is stated: 
```
Only use catch(...) directly inside of main(), and then only to clean up resources before terminating the program
```
The spirit behind this rule is often `catch(...)` ends up destroying valuable information about an exception condition, and even worse, if there's no rethrow it can allow a program to muddle along in an undetected bad state rather than failing loudly (which is preferable). The fact that in the catch-all blocks here in dfmodules other exceptions get thrown ("Unknown exception thrown by HDF5") takes away a lot of the sting of a `catch(...)` in that the muddle-along scenario is avoided. I leave it to your best judgement as to whether this approach is sufficient, or if there are scenarios where we may lose valuable information as to something going wrong; if it's the former then I approve of this PR.

jcfreeman2:
Also, concerning the linter warnings along the lines of "these functions are too long": the linter considers functions longer than 80 lines as being in danger of becoming "Swiss Army Knife" functions, or at least insufficiently modular. Probably this isn't an issue with the functions we get warnings for; if so just slap a `// NOLINT` after the closing brace of the long functions.

mroda88:
Hi @jcfreeman2 I had a look. All the `catch(...)` are re-throwing so I think they are ok as they are. Since they are relate with disc operations I put them there ages ago in order to protect against any inconvenience, especially since we don't know if these exception will be thrown into the main thread of the application. So there might not be a main to catch those. 
Specifically the re-throwing is in the form a ERS exception, so it shall give the loudest possible message if CCM does its job. 

About the do_work(). That function is in a way special, as it is the main loop that operates for that specific module. It is obviously a complex function, so I would be in favour of the `// NOLINT` solution. Yet I have identified a few blocks that could be turned into separate functions and since we are just about over 80 lines (88) we can fix it. What do you prefer?

eflumerf:
I added NOLINT to the HDF5DataStore catch(...) along with a comment about how we're rethrowing anyway. I also added NOLINT to the two long do_work functions, as they are "special".

jcfreeman2:
Sounds good to me; I'm fine with the pull.

PR #81, opened 2021-05-21
Made changes needed to convert to using Info Structs from opmonlib

bieryAtFnal:
Alessandro mentioned the need to make these changes several days ago in the daq-integration Slack channel:

- `opmonlib has now custom templates for Info structs. The immediate purpose is to standardise the definition of class_name but it’s likely they will be used for further customisation. All packages targeting  2.6 should switch to using those. Will prepare some instructions early next week. In the meantime, listrev is probably the simplest example to follow:
  https://github.com/DUNE-DAQ/listrev/pull/19/files`

Here is the link:  [listrev sample](https://github.com/DUNE-DAQ/listrev/pull/19/files)

The changes in this PR are my attempt to apply the example changes to the _dfmodules_ repo.  

I have run tests that show that A) the _dfmodules_ code compiles, B) the generated JSON for the _dfmodules_ jsonnet files is identical to what was produced before the change, and C) I see the expected opmon quantities being produced from the _dfmodules_ code.


mroda88 (approved):
I compiled and run with this new branch.
I can also confirm that the  metrics have the same structure.

PR #82, opened 2021-05-26
Added sanity check on reasonably-sized number of ticks when trying to…

bieryAtFnal:
… convert to a human-readable datetime string in hdf5_dump.py


bieryAtFnal:
When I first tried to use hdf5_dump.py on fake PDS data, the large values for the timestamps in the trigger and fragment headers caused the script to abort...

dingp (approved):
LGTM

PR #83, opened 2021-06-11
Sample integration tests

bieryAtFnal:
I created some sample integration tests based on Phil's integrationtest package.


bieryAtFnal:
There are some notes in the README file in the integtest directory that try to explain how to run the tests.

philiprodrigues:
These look sensible to me, and I was able to successfully run `four_process_empty_run_test.py` locally. I've updated `integrationtest` so that it uses `nanorc` from `$PATH` if `--nanorc-path` is not specified, and so that the data file sanity check doesn't assume that the trigger numbers in the file start at 1, which fixes the two problems you mention.

As for next steps, I will schedule a a discussion with software co-ordination about whether/how to include `integrationtest` in the DAQ software release. In the meantime, I don't see any harm in merging this PR, although maybe you prefer to wait until `integrationtest` is included in a release, or at least we've decided on its status.

PR #84, opened 2021-06-15
Trigger Record timeout

mroda88:
I'll convert in proper PR once I can test it


bieryAtFnal:
Marco, I thought a little more about the need to specify the timeout in clock ticks, and that struck me as a little odd, since I thought that this test would be based on the system time of the computer on which the TRBuilder is running.

Looking at the code, I see that the timeout is compared to a difference of _data_ times instead of _system_ times.  I understand your motivation for doing it that way, but making this test depend on the presence of new data/TriggerDecisions does not seem like it will give the behavior that we want (or at least the behavior that I was expecting).  

I was expecting that an incomplete TriggerRecord would be sent downstream N system clock seconds after either [the system time when the TRB received the TD] OR [the system time when the first fragment was received from the Readout processes] (if I specified N in the configuration).
Isn't it true that the current code would send it downstream at the larger of N seconds and the arrival of the next TD after N?  So, if I set the fake trigger rate to one per minute, I'll only get incomplete TRs sent every minute, instead of 10 seconds after they became stale?

bieryAtFnal:
Yes, please, let's move in the direction of having this test **not** be data driven.  

As you suggest, this direction will require some changes.  One will be keeping track of the system time of the arrival of each TD (for later comparison), and another will be the creation of a thread to periodically do the checking.  That's fine and certainly worthwhile.

Thanks!

mroda88:
ok. Couple of questions though.
1) Why another thread? Is it not enough to have the standard thread to check this? I agree that the thread might go to sleep and hence the timeout might be delayed, is it so important to be precise? A second thread implies either a different map or a mutex and at our low rate, both have much worse performances. It's doable but I prefer to have dedicated discussion first then. 
2) once we add the bookkeeping of the time, it will be easy to implement a metric that has been mentioned a while ago: the average time to process a TR. Is it ok if we add that as well in the PR?

bieryAtFnal:
Ah, sorry, I was (incorrectly) picturing the main thread being unable to do this check in addition to its other work.  So, yes, you're right, no extra thread is needed.

Adding metric(s) as part of this PR sounds fine to me.

mroda88:
Hi @bieryAtFnal I think I implemented what we discussed. I only tested rapidly, and it seems to work. I'll test more tomorrow, but I think you can start having a look.

bieryAtFnal:
Marco, I ran some tests and the results look good.  For example, I enabled the BOOKKEEPING trace level and saw the timed-timed out TRs dropping off the list at the expected time.  

Following up on our discussion of whether the "old" metrics still make sense...  In the test that I ran, I had the timeout value set to a smaller number than the "old" value, so I never saw any non-zero "old" metric values.  And, as you said, having two 'timeout' values will probably be confusing.  So, I'm voting that we get rid of the "old" metrics.

mroda88:
I think this is ready to go. I performed quite some tests, stressed the system and found it stable, and if not I'm certain it was not TRBuilder fault. 
I also added all the metrics that can reflect error conditions in order to avoid to dig into the logs.

bieryAtFnal:
I also ran some tests with the latest code on this branch (trigger rate was 1 Hz)

1. ran test with default conditions and one of two (readout) DataLinkHandlers not sending any data
   * saw 10 TRs (with only one Fragment each) in the data file, no matter how long the run was (30 or 60 seconds), which made sense
   * looking at the opmon metrics…
      * "average_millisecond_per_trigger" was always -1, presumably because no trigger records were ever completed
      * "timed_out_trigger_records" was always zero, presumably because the default timeout value is set to zero (no timeout)
      * what are the "deleted_fragments" and "deleted_requests" metrics?
         * do these measure the number of items that are popped off the relevant queues, but are unexpected (because they don’t match the map of expected GeoIDs or the list of partial TRs)?
         * if so, I wonder if "dropped" or "abandoned" might be more descriptive than "deleted"
   * I added a TLOG() message to indicate when incomplete TRs are sent downstream during Stop, and with those messages, I verified that the incomplete TRs were sent downstream at end-run time, as expected.  (Is it OK if I commit this change?)

2) TRBuilder.cpp appears to need clang format.  I can do this, if we want.

3) ran test with the TR timeout set to 2.5 seconds (2500 msec).  Saw the incomplete TRs being sent downstream after the timeout, as expected. In the opmon stats, the "average_millisecond_per_trigger" was approximately 2500, which made sense.  The "timed_out_trigger_records" metric increased over the course of the run, as expected, and the final number matched the number of TRs in the data file.  The values for “fragments” and “trigger_decisions” were typically 2 during the run, which made sense.  (However, I wonder if we might come up with better names for those metrics.  I believe that these are the number of fragments and TDs that are currently stored in the Book, but it is hard to tell from the metric names that they aren’t something else, for example, the total number of fragments or TDs seen so far.)

4) ran test with the TR timeout set to 15 seconds. Saw incomplete TRs timing out during the 60-second run **and** being sent downstream at Stop time, as expected.  “average_millisecond_per_trigger” was around 15000; “fragments” and “trigger_decisions” were around 9-10.  40 TRs were written to the raw data file in a 60-second run, which seemed right.

bieryAtFnal:
The latest changes look good to me.  
One last question: do we think that pending_fragments should include the count (sum) of every fragment that was missed during the run, or the instantaneous number of fragments that are needed to complete the TRs that are in the Book at the moment?  It looks like the former is what we have now.  In a test in which events timed out, I think that I was initially expecting the latter.  par t of that was the ability to know how many DataLinkHandlers were misbehaving at the current moment...

bieryAtFnal:
Sorry, yet another "last question"...  (Things are looking quite nice, though!)

In my tests with the latest code, all of the metrics have the values that I expect (and I appreciate the new names).  

In a run with some number of timed-out incomplete TriggerRecords, and some number of incomplete TriggerRecords sent downstream at Stop time, the numbers of _lost_fragments_ and _timed_out_trigger_records_ have the expected values, given a very literal interpretation of their names.  However, they didn't have the same numeric value at the end of the run (I had disabled the sending of Fragments from **one** of the DataLinkHandlers, so they might have.)  The reason for the difference is that _last_fragments_ must be counting Fragments that are lost for any reason (TR time out or abandoned at Stop time), whereas _timed_out_trigger_records_ is probably only counting TR timeout.  Do we care?

mroda88:
This is exactly why, once we merge, we (I, I suspect) need to write down some explanations and possible cases to make it clear what metric values represent in terms of situations. 
In the specific case of your test, I don't care but probably we need to ask someone else's opinions. There are so many cases that it's hard to track all the possibilities. The question is: are the present metrics describing all the useful situations?

One thing that might be useful, is having a little zoom session, where we try to run multiple tests and we decide if we have enough information out of the metrics. Maybe from a common machine or system where we can check simultaneously the system's responses.

bieryAtFnal:
_The question is: are the present metrics describing all the useful situations?_

Yes, what we have right now looks good.  So, let's go ahead with the merge and move on to the documentation that you describe.

PR #88, opened 2021-07-12
Run clang-format and fix some linter issues

eflumerf:


jcfreeman2:
Formatting is correct and it's of course nice to see `#include <stdlib.h>` switched to `#include <cstdlib>`

eflumerf:
This PR is stale and the clang-format and linter steps should be repeated.

PR #89, opened 2021-07-15
Move `data_file_checks` into dfmodules

philiprodrigues:
This PR moves the `data_file_checks` python module into dfmodules (from `integrationtest`). This is maybe a better home (nearer the code that is writing the data files), and it allows me to check one of the requirements for moving `integrationtest` into DUNE-DAQ, which was that we can use test functions from other modules. A test that uses the `data_file_checks` from `dfmodules` is in `minidaqapp` in the philiprodrigues/integration-test-use-dfmodules branch


bieryAtFnal:
Phil, I made several updates to this branch

1. I re-merged _develop_ into this branch to pick up the latest changes from _develop_
2. I added `requirements.txt, setup.cfg, and setup.py` files.  This is the change that I'm not sure that I should have made, but it seems like this was necessary to get the pytests to find the dfmodules.data_file_checks.  Please let me know if I should back this out.
3. I renamed and modified the dfmodules/integtest/README file.

Something else that I wasn't sure about:  I needed to run `pip install .` after adding the files listed in #2 above.  Was that the right thing to do?

With those changes, the tests in minidaqapp/integtest and dfmodules/integtest run.   They complain about unexpected Fragment sizes, but I/we can fix that in a separate step.

bieryAtFnal:
Phil and I had a chance to talk about this.  (He expected that I wouldn't need to do the 'pip install' stuff for _dfmodules_).

We realized that I had used a software area based on a Nightly build (07-15), whereas he had been using a software area based on dunedaq-v2.6.0.  I confirmed that things work when I use a 2.6 base release, and Phil found that the dfmodules python files are not being copied into build/dfmodules/python/dfmodules when using the Nightly as the base release.

One difference that we know of between nightly-07-15 and dunedaq-v2.6.0 is a newer version of daq-cmake.  I can try some tests with different versions of that package...

bieryAtFnal:
I've confirmed that the different behavior has to do with the new version of daq-cmake.  (I used the 2.6 daq-cmake in a software area based on the 07-15 Nightly build and the problem went away.)
And, vice-versa.  Adding the new daq-cmake to a software area based on 2.6 causes the problem to appear.
("the problem" may not be a problem, per se.  I just mean that the dfmodules python files are not installed in build/dfmodules/python/dfmodules)

bieryAtFnal:
Maybe we need to switch to the new daq-buildtools when using the new daq-cmake.  I will try that.

bieryAtFnal:
Using the new daq-buildtools with the new daq-cmake doesn't seem to help.  The build/dfmodules/python/dfmodules directory still does not get created.  
(Another caveat:  this may be "by design" with the new daq-cmake; Phil is checking with John.)

philiprodrigues:
We were running into a couple of overlapping changes with the nightly: the nightly includes a new `daq-cmake` which requires the use of the HEAD of `daq-buildtools`. The new `daq-buildtools`/`daq-cmake` combination runs everything out of the install directory, not the build directory, which is why we didn't see the python files in the build directory - they _are_ present in the install dir. But the HEAD of `daq-buildtools` was setting `$PYTHONPATH` slightly wrong, so the files in the install directory weren't being found. John fixed this in https://github.com/DUNE-DAQ/daq-buildtools/pull/152 . With the branch from that PR, I can successfully use `data_file_checks.py` from `dfmodules` without needing to create `setup.py` etc and `pip install`

bieryAtFnal (approved):
All of this looks good to me. However, I have made a number of changes myself.
So, I think that it would be great if Phil could look at the changes that I have made.
And, maybe we ask Eric to take a look, too, if he has some time today.
From my perspective, this is ready to be merged.

eflumerf:
My only comment is that the diff shows that some parameters could probably be either declared early, parameterized, or calculated. (Trigger window before/after, sleep times between resume and stop, expected Fragment sizes).

philiprodrigues:
The code all looks sensible to me, and I was able to run all of the tests with the N21-07-22 nightly, which includes `integrationtest` in the python venv

PR #85, opened 2021-06-22
Use sequence number - first step

mroda88:

PR #90, opened 2021-07-15
Refinements

mroda88:
Following today's discussion, I added the required metrics to cross check with other modules and introduced a check on the run number for incoming trigger decisions. 
Also linted and clang formatted. 
The PR is based on PR #85 so first let's merge that one.


include/dfmodules/StorageKey.hpp
```diff
@@ -48,15 +48,20 @@ class StorageKey
...
+  }
 
   int get_run_number() const;
   int get_trigger_number() const;
   DataRecordGroupType get_group_type() const;
   int get_region_number() const;
   int get_element_number() const;
 
+  int m_this_sequence_number;
+  int m_max_sequence_number;
```
eflumerf, line 63: We could use
```CXX
int m_this_sequence_number{0};
int m_max_sequence_number{0};
```
and remove the lines from the constructor.

Are these members supposed to be public?

plugins/HDF5KeyTranslator.hpp
```diff
@@ -75,6 +75,9 @@ class HDF5KeyTranslator
     trigger_number_string << m_data_record_params.trigger_record_name_prefix
                           << std::setw(m_data_record_params.digits_for_trigger_number) << std::setfill('0')
                           << data_key.get_trigger_number();
+    if (data_key.m_max_sequence_number > 0) {
```
eflumerf, line 78: Are negative values reserved for error cases? Should these be an unsigned type?

plugins/TriggerRecordBuilder.hpp
```diff
@@ -256,24 +271,35 @@ class TriggerRecordBuilder : public dunedaq::appfwk::DAQModule {
...
-  mutable std::atomic<uint64_t> m_data_request_width= {0};      // in between calls
-  mutable std::atomic<metric_counter_type> m_received_trigger_decisions = {0}; // in between calls
-  mutable std::atomic<metric_counter_type> m_generated_data_requests = {0}; // in between calls
+      0}; // in between calls
+  mutable std::atomic<uint64_t> m_trigger_record_time = {
+      0}; // in between calls // NOLINT
+  mutable std::atomic<uint64_t> m_trigger_decision_width = {
+      0}; // in between calls // NOLINT
+  mutable std::atomic<uint64_t> m_data_request_width = {
+      0}; // in between calls  // NOLINT
```
eflumerf, line 298: These NOLINTs have lost their effect

plugins/TriggerRecordBuilder.hpp
```diff
@@ -45,53 +45,65 @@ struct TriggerId {
 
   TriggerId() = default;
 
-  explicit TriggerId(const dfmessages::TriggerDecision& td, 
-		     dataformats::sequence_number_t s = dataformats::TypeDefaults::s_invalid_sequence_number)
-    : trigger_number(td.trigger_number)
-    , sequence_number(s)
-    , run_number(td.run_number)
-  {
+  explicit TriggerId(const dfmessages::TriggerDecision &td,
```
eflumerf, line 48: Is this struct large/complex enough to move into its own file?

mroda88: Good question. I would vote for no, as it's really only internal. 
But I have no strong opinions. It could be moved in the include of this package if we think it's worth it.

eflumerf:
The linter doesn't like `TriggerRecordBuilder::create_trigger_records_and_dispatch()` as it is too long. Looking at the code, I see comments that indicate that a certain amount of refactoring could be done (e.g. "// loop over component in trigger decision").

My usual complaint that there are lines and logic added without corresponding unit test changes also applies.

eflumerf (approved):
Changes look okay, builds, and unittests work. I have a few higher-level comments that don't need to be addressed to merge the code.

eflumerf (approved):
LGTM

PR #92, opened 2021-07-26
Very preliminary step for TPSet stream writing

bieryAtFnal:
The intention of the changes on this branch are to (somewhat crudely) demonstrate how we might write the TPSet stream to disk.

1. There is a new TPSetWriter plugin that will definitely need to change once we learn more about the format and contents of the TPSet stream.  For example, at the moment, there isn't really any "building" of common time windows.  It just writes the TPSets as they arrive.
2. As we've decided in the developer meetings, we will simply write fragment-only Fragments for v2.8.  So, the TPSet serialization code is commented out in TPSetWriter.
3. The configuration code in dfmodules/python/dfmodules/tpset_demo should be a reasonable start for a "local" test.  This config just uses a fake TPSet creator.
4. Before we take the next steps, we will need to talk about what tools we want to use in the future. readout::BufferedFileWriter was used in this example, but there are pros and cons to that choice.
5. etc, etc.  (There are probably other caveats that I'm forgetting...)


mroda88 (approved):
Configuration works and I managed to run and obtain expected output. 
I only checked that the final output has a pattern that resemble a Trigger Record header. 
I'll have a look at the code and validate the output more carefully tomorrow.

mroda88 (approved):
I had a look at the code as well, all seems ok to me

PR #93, opened 2021-07-27
Fgrotsch/fake data prod resurrection

floriangroetschla:
Resurrection of the old fake data producer with some additional features like configurable latency, more realistic fragment size and opmon info.


bieryAtFnal (approved):
Looks good to me.

PR #94, opened 2021-08-02
Changes related to better handling of a full disk when trying to write data

bieryAtFnal:
The original impetus for these changes was when Giovanna noticed that the Dataflow App would crash when the output disk filled up.  In addition to handling that condition, these changes add a few other improvements.

The changes:

- ERS Issues in DataStore.hpp, DataWriter.hpp/cpp, and HDF5DataStore.hpp were modified somewhat.  This is to bring them in line with best practices (e.g. not having the word "Error" in the name of the Issue), to provide the right level of detail in each place in the call stack, and to provide finer-grain control over the reaction to certain exceptional conditions (e.g. "can/should the operation be retried, or would that be fruitless?")
- Exception handling was added to the Stop transition (DataWriter and HDF5DataStore).
- An explicit free-space check was added to the HDF5DataStore write operation.  This will cost us a little bit in terms of performance, but I found that the original crash was happening inside the HDF5 code (e.g. HDF5_Flush), and there appeared to be no way to prevent that.  So, the right reaction seemed to be to never call the HDF5 write routine(s) when the disk is full.
- Retries of write operations were added to DataWriter.  The DW code checks if the exceptional condition is something that a retry might fix, and if so, it retries it, as long as a Stop command has not been received.
   - This has the very practical effect of avoiding data loss.  In the earlier code, if an exception was encountered during a write operation, the current TriggerRecord was dropped on the floor and data taking continued.  Now, the write operation will be retried until either the run ends or the write succeeds (because someone cleared up some disk space).
   - If a run ends while the DataWriter is stuck retrying the write of a given TR, it will give up on that TR and attempt to write all of the TRs in its input queue that are waiting.  For each of those, it tries just once and then goes on.  If the write attempt for each TR fails, an error is reported.

Configuration parameters were added so that users can specify the following values:

- how much of a free-space safety factor should be required when writing each TriggerRecord.  (The code enforces a minimum of 1.1, i.e. 10% overhead.)  Default value is 5.
- what minimum and maximum retry intervals to use.  And, the compounding factor to use on each iteration.  (The idea was to implement something like an exponential back-off.).  So, for a write_retry_time_increase_factor value of 2, the wait time between retries will double each time.  The code enforces a minimum value for the min_write_retry_time_usec of 1 usec.  Defaults values are 1 msec, 1.0 second, and factor of 2.

Tests were done in which a disk was filled up during one run, that run was allowed to continue for a little while, then it was ended. and a second run was started without freeing up any space.  The following behaviors were validated:  the absence of a crash when the disk filled up, the retries of the writing of the next TR beyond the "full" condition, the draining of the DataWriter input queue on Stop, the failure of the second run to start because the disk was full.  
Tests were also done in which a full-disk condition happened during a run, that run was allowed to continue, and some free space was created.  In that case, data-writing resumed.

Most of my tests were done on a seldom-used (at the moment, at least) server at NP04 (np04-srv-002), and I would temporarily fill up the /tmp disk.
Here are some commands that might be useful reference:

- `python -m minidaqapp.nanorc.mdapp_multiru_gen --host-ru localhost -d $HOME/.biery/dunedaq/frames.bin -o . -s 10 -b 100000 -a 100000 -t 5 -o /tmp/biery full_disk_test`
- `let run1=930; runduration=100; waitAfterStop=11; let run2=$run1+1; nanorc full_disk_test boot init conf start ${run1} wait 2 resume wait ${runduration} pause wait 2 stop wait ${waitAfterStop} start ${run2} resume wait 10 stop scrap terminate; df -h /tmp ; rm -fv /tmp/biery/*${run1}* ; df -h /tmp`

In order to see all of the retry-failed error messages in the log_dataflow_<portno>.txt file, I would change the thresholds for ERS messages in the configuration boot.json file.  For example:

```
(dbt-pyvenv) [np04daq@np04-srv-002 full_disk_test]$ diff boot.json boot.json~
25c25
<         "DUNEDAQ_ERS_ERROR": "erstrace,throttle(1000,0),lstderr",
---
>         "DUNEDAQ_ERS_ERROR": "erstrace,throttle,lstderr",
27c27
<         "DUNEDAQ_ERS_INFO": "erstrace,throttle(1000,0),lstdout",
---
>         "DUNEDAQ_ERS_INFO": "erstrace,throttle,lstdout",
29c29
<         "DUNEDAQ_ERS_WARNING": "erstrace,throttle(1000,0),lstderr",
---
>         "DUNEDAQ_ERS_WARNING": "erstrace,throttle,lstderr",
```

### DIFF
--- a/include/dfmodules/DataStore.hpp
+++ b/include/dfmodules/DataStore.hpp
@@ -64,17 +64,6 @@ ERS_DECLARE_ISSUE(dfmodules,               ///< Namespace
 )
 /// @endcond LCOV_EXCL_STOP
 
-/**
- * @brief A generic ERS Issue for DataStore writing failure
- * @cond Doxygen doesn't like ERS macros LCOV_EXCL_START
- */
-ERS_DECLARE_ISSUE(dfmodules,                                     ///< Namespace
-                  DataStoreWritingFailed,                        ///< Type of the Issue
-                  "Failed to write data using  " << plugin_name, ///< Log Message from the issue
-                  ((std::string)plugin_name)                     ///< Message parameters
-)
-/// @endcond LCOV_EXCL_STOP
-
 namespace dfmodules {
 
 /**
--- a/include/dfmodules/DataStore.hpp
+++ b/include/dfmodules/DataStore.hpp
@@ -53,7 +53,7 @@
 namespace dunedaq {
 
 /**
- * @brief A ERS Issue for DataStore creation failure
+ * @brief An ERS Issue for DataStore creation failure
  * @cond Doxygen doesn't like ERS macros LCOV_EXCL_START
  */
 ERS_DECLARE_ISSUE(dfmodules,               ///< Namespace
@@ -62,6 +62,28 @@ ERS_DECLARE_ISSUE(dfmodules,               ///< Namespace
                                                 << conf,           ///< Log Message from the issue
                   ((std::string)plugin_name)((nlohmann::json)conf) ///< Message parameters
 )
+/// @endcond LCOV_EXCL_STOP
+
+/**
+ * @brief An ERS Issue for DataStore problems in which it is
+ * reasonable to retry the operation.
+ * @cond Doxygen doesn't like ERS macros LCOV_EXCL_START
+ */
+ERS_DECLARE_ISSUE(dfmodules,
+                  RetryableDataStoreProblem,
+                  "Module " << mod_name << ": A problem was encountered when " << description,
+                  ((std::string)mod_name)((std::string)description))
+/// @endcond LCOV_EXCL_STOP
+
+/**
+ * @brief An ERS Issue for DataStore problems in which it is
+ * not clear whether retrying the operation might succeed or not.
+ * @cond Doxygen doesn't like ERS macros LCOV_EXCL_START
+ */
+ERS_DECLARE_ISSUE(dfmodules,
+                  GeneralDataStoreProblem,
+                  "Module " << mod_name << ": A problem was encountered when " << description,
+                  ((std::string)mod_name)((std::string)description))
 /// @endcond LCOV_EXCL_STOP
 
 namespace dfmodules {
--- a/plugins/DataWriter.cpp
+++ b/plugins/DataWriter.cpp
@@ -291,8 +291,8 @@ DataWriter::do_work(std::atomic<bool>& running_flag)
 
         bool should_retry = true;
         int retry_wait_usec = 5000;
-	do {
-	  should_retry = false;
+        do {
+          should_retry = false;
           try {
             m_data_writer->write(data_block_list);
             ++m_records_written;
@@ -317,8 +317,7 @@ DataWriter::do_work(std::atomic<bool>& running_flag)
                                           trigger_record_ptr->get_header_ref().get_run_number(),
                                           excpt));
           }
-        }
-	while (should_retry && running_flag.load());
+        } while (should_retry && running_flag.load());
       }
     }
 
--- a/plugins/DataWriter.cpp
+++ b/plugins/DataWriter.cpp
@@ -99,6 +99,12 @@ DataWriter::do_conf(const data_t& payload)
   TLOG_DEBUG(TLVL_CONFIG) << get_name() << ": initial_token_count is " << conf_params.initial_token_count;
   TLOG_DEBUG(TLVL_CONFIG) << get_name() << ": data_storage_prescale is " << m_data_storage_prescale;
   TLOG_DEBUG(TLVL_CONFIG) << get_name() << ": data_store_parameters are " << conf_params.data_store_parameters;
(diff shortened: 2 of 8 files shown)


mroda88:
Ok, first test. As a beginning I tried to obtain the error. I run on the same machine you are suggesting, but I'm using the graphana to monitor the situation so I didn't change the ERS related variables. 

Screenshot from 2021-08-02 17-43-30

I did a screenshots of the errors and they seem loud enough for me and very visible. 

The applications didn't die and the rest of the system seems to behave accordingly: we inhibit triggers and the rest of the metrics seems to reflect correctly the behaviour. 
Of course the run is effectively stopping, but it seems this is the desired behaviour. 

I'll try to have a look at the code.

mroda88:
Performance test. 
With the following configuration
```
python -m minidaqapp.nanorc.mdapp_multiru_gen --host-ru localhost  -b 100000 -a 100000 -t 300 -c 5000 -o /tmp/maroda --opmon-impl cern --ers-impl cern disk_performance
```
I managed to reach a condition in which the TRB was not able to push more TRs into the queue as the DataWriter was too busy writing, mainly we had a lot of ERS warnings like
```
trigger_record_q: Unable to push to output queue "trb" within timeout period (timeout period was 100 milliseconds)
```
In this condition the peak of the metric
```
SELECT derivative(mean("bytes_output"), 1s)  / 1073741824 FROM "dunedaq.dfmodules.datawriterinfo.Info" WHERE ("source_id" = 'maroda_test_dataflow.datawriter') AND $timeFilter GROUP BY time($__interval) fill(null)
```
gives us the maximum writing speed. 
I did two tests: one with this branch and one with develop. The difference was small, but visible:
Screenshot from 2021-08-03 15-37-03
The first peak was 1.15 GiB/s, while the second was 1.25 GiB/s.

I believe this is the price we pay to add the check. Considering that to reach this effect I had to go at 300 Hz, I believe this is not a problem.

mroda88:
I run the same test on a larger disk and I see a plateau after the peak. That is much lower but I see a similar reduction in efficiency:
with disk check 0.145 GiB/s
without disk check 0.152 GiB/s
around 5 %

mroda88 (approved):
I think the code looks well and behaves well. 
I managed to get into the error condition (disk full) many times and managed to stop the run, clean the disk and start a new run without having to reboot the applications or the run control.

Having said that, there a few issues related to the system that I noticed. They are not a fault of this PR, but this PR allowed the condition to be observable, so I'll make a not here. 

- when in the error condition, nanorc always times out the stop command response wait. This might scare an inexperienced user. But indeed the run and the applications are not lost. So this is a minor concern.
- When in the error condition, the actual stopping of the run, meaning the release of the threads of all the modules in the dataflow, can take ages (several seconds in fact). This is mostly on the TRB side. This is predictable: it is the time necessary for the TRB to dispose of the TRs in its buffer, every TR requires the queue_timeout time to complete the operation, and of course the TRs are lost as they don't reach the disk. How concerned are we of both these facts?

mroda88:
If we are concerned about the slow stop, #95 will fix it

bieryAtFnal:
Marco, I re-did some of my tests, and I don't see a slow stop.  For example, in one of my tests, the following processes reported their stop times:

- HSI: 20:16:36.340
- Trigger: 20:16:34.928
- Dataflow: 20:16:36.340

I presume that you are not worried about 1.4 seconds, so I wonder what is different about your test setup compared with mine...

Naively, i would expect the stop to happen fairly quickly.  In my test, the number of tokens is 10 and the size of the trigger_record_q is 100.  So, when the system is stuck in a situation in which the output disk is full, all of the 10 TriggerRecords that are 'in flight' would be in the queue between the DF TRB and the DF DataWriter, right?  So, when the Stop command arrives, the DataWriter just tries to write each of the 10, fails, and gives up, I believe.

Stated another way, I'm not sure how a change in the TRBuilder code would affect the stop time in the situation in which all 10 TRs should already be downstream of the TRBuilder.  That probably just means that I'm confused about the different tests that we are running.  Maybe you typically run with a large number of tokens and/or a small queue size between the TRB and the DW...
Thanks,
Kurt

mroda88:
Hi Kurt, you are correct: the delay is due to the high number of tokens I was using in this specific test. Still it is a problem, because it is a situation that might happen. Here are the details of what is happening. 

At the end of my runs there were almost a thousand TR in the book. In addition, since the disk was full, at the stop, emptying the book took 1000*100ms = 100 s. This is because every TRs that should be send to DataWriter will have to timeout before the TR is released (i.e. deleted).

This is not a delay in the propagation of the stop signal, it is a delay in the completion of the transition itself, one that the appfwk is not able to handle. So the stop is reached and confirmed via run control, but in reality the thread is still working, making it a very dangerous situation for the state machine. 
You can see the moment when the stop is actually completed only via the metrics. 

I hope this small description is enough for now, to explain what happens. Tomorrow I'm happy to answer to all doubts.

PR #95, opened 2021-08-03
Fast stop when disk is full

mroda88:
This will fix the slow stop when disk is full


bieryAtFnal (approved):
These changes look good to me.
I ran a test in which I artificially added sleep to the DataWriter and increased the number of tokens beyond the size of the trigger_fragment_q (which is between the DF TRB and the DataWriter).  That triggered the behavior in this code, and I saw the ERS messages and metrics that I expected.

PR #96, opened 2021-08-04
Updated cernbox location of frames.bin file in integtest README

bieryAtFnal:
This is based on the new frames.bin file that Florian has provided.


carloschavez:
@bieryAtFnal ,  I have tested that the location of the frame file is correct. And ran successfully the pytest  four_process_quick_test.py

PR #100, opened 2021-08-17
Tweaks to the HDF5 Group name for TP_APA

bieryAtFnal:
Here is the request from Giovanna:

```
GROUP "TPC” {
  GROUP “TP_APA000" {
    DATASET "Link02" {
    DATASET "Link03" {
  } } }
```

The code changes associated with this PR implement this requested naming.

To test, we can use
* `python -m minidaqapp.nanorc.mdapp_multiru_gen --host-ru localhost -d $PWD/frames.bin -o . -s 10 --enable-software-tpg mdapp_4proc_swtpg`

### DIFF
--- a/plugins/DataWriter.hpp
+++ b/plugins/DataWriter.hpp
@@ -93,8 +93,13 @@ private:
   using geoid_system_type_t = dataformats::GeoID::SystemType;
   using key_group_type_t = StorageKey::DataRecordGroupType;
   std::map<geoid_system_type_t, key_group_type_t> m_system_type_to_group_type_mapping;
-  key_group_type_t get_group_type(geoid_system_type_t system_type)
+  key_group_type_t get_group_type(geoid_system_type_t system_type, dataformats::FragmentType frag_type)
   {
+    // 16-Aug-2021, KAB: ugly hack
+    if (system_type == geoid_system_type_t::kTPC && frag_type == dataformats::FragmentType::kTriggerPrimitives) {
+      return key_group_type_t::kTPC_TP;
+    }
+
     if (m_system_type_to_group_type_mapping.size() == 0) {
       m_system_type_to_group_type_mapping[geoid_system_type_t::kTPC] = key_group_type_t::kTPC;
       m_system_type_to_group_type_mapping[geoid_system_type_t::kPDS] = key_group_type_t::kPDS;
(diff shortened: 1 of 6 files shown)


plugins/DataWriter.hpp
```diff
@@ -93,8 +93,13 @@ class DataWriter : public dunedaq::appfwk::DAQModule
   using geoid_system_type_t = dataformats::GeoID::SystemType;
   using key_group_type_t = StorageKey::DataRecordGroupType;
   std::map<geoid_system_type_t, key_group_type_t> m_system_type_to_group_type_mapping;
-  key_group_type_t get_group_type(geoid_system_type_t system_type)
+  key_group_type_t get_group_type(geoid_system_type_t system_type, dataformats::FragmentType frag_type)
```
mroda88, line 96: instead of adding objects here I would have used a single input, to be the whole geoid itself. 
I would make the whole interface more flexible

mroda88 (approved):
I made a small comment, I don't know if we want to fix it or not. 

Anyway, the code works well I tested it and for the test suggested in the PR, for a TR I see the following components:
```
TriggerRecordHeader
TPC/APA000/Link00
TPC/APA000/Link01
TPC/TP_APA000/Link02
TPC/TP_APA000/Link03
Trigger/Region000/Element00
Trigger/Region000/Element01
```
If this is correct we are good to go. 
What is a bit odd here is the different link numbers for the APA and TP_APA. Is it correct?

bieryAtFnal:
Hi Marco,
Yes, at the moment, we have different link numbers for APA and TP_APA.  That is a simplification that is taken in the configuration/mapping of components.

For the function arguments, the FragmentType isn't part of the GeoID, and that is why I included it separately.  As you suggest, there will certainly be improvements that we can make to the handling of HDF5 characteristics, but let's defer those for now.

In that regard, you may recall that we have two HDF5 Attributes in each raw data file that can be used to help track when things change:

```
GROUP "/" {
   ATTRIBUTE "data_format_version" {
      DATATYPE  H5T_STD_I32LE
      DATASPACE  SCALAR
      DATA {
      (0): 2
      }
   }
   ATTRIBUTE "operational_environment" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
      DATA {
      (0): "swtest"
      }
   }

```
I can imagine that even these may change once we have a change to talk about HDF5 topics, but at least we'll be able to notice when something changes.
Thanks,
Kurt

PR #101, opened 2021-08-19
v2.8 integtest updates

bieryAtFnal:
Eric and I have been fixing and adding to the integration tests in the dfmodules/integtest directory.

The purpose of this PR is to give us a place to start commenting on changes and eventually get them merged to _develop_.

At the moment, I believe that all of the tests are working, but we should note that there are changes in other packages (e.g. _nanorc_ and _integrationtest_) that will improve the behavior of these tests once those changes become available.


bieryAtFnal (commented):
Eric and I have tested these changes successfully.
And, they don't affect the primary workings of the package.
And, they are definite improvements on what is part of the previous tag.
So, I'm going to go ahead and merge them.

PR #102, opened 2021-09-22
Remove username from data filenames

bieryAtFnal:
Alessandro recently told me that the use of usernames in the raw data filenames produced by the HDF5DataStore causes problems when running our daq_applications inside containers because the USER env var may not be set.

Unfortunately, I had not been checking that the `getenv` call was returning a non-null pointer, but given the trend of our work toward non-shared environments, I have removed the use of $USER in the raw data filenames.

### DIFF
--- a/plugins/HDF5DataStore.hpp
+++ b/plugins/HDF5DataStore.hpp
@@ -114,9 +114,7 @@ public:
     : DataStore(conf.value("name", "data_store"))
     , m_basic_name_of_open_file("")
     , m_open_flags_of_open_file(0)
-    , m_username_substring_for_filename("_UnknownUser")
     , m_timestamp_substring_for_filename("_UnknownTime")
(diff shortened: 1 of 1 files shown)


eflumerf (approved):
Looks good. Builds and runs and produces output files with username portion deleted.

What about using the partition name?

bieryAtFnal:
Good idea! (partition name)
I have mixed feelings about using it, though.  On the one hand, I like it because it can probably be determined reliably and would likely be unique at a given point in time.  On the other hand, I think that we should give as little indication of partition number/name to downstream users because we don't want them to think that partition number/name is a useful indicator of anything; it shouldn't be - all partitions should be equivalent.  Lastly, I'd like to see how several investigations/discussions (containers; feedback from offline on what they want to see) play out before adding back something that might not be needed.

So, let's move forward with only timestamp, for now, and see how things go.

PR #103, opened 2021-09-22
integtest event count

bieryAtFnal:
In an attempt to provide additional validity checking in several of the integration tests that we have in dfmodules/integtest, I've added validation of the expected number of trigger records.

The test generally succeeds, but I suspect that there may be some computers on which the event rate is low (because of a lack of sufficient resources). 

However, I'm submitting these changes as a pull request to get them reviewed by others, and if approved, to make them available for others to use in testing.

### DIFF
--- a/integtest/four_process_disabled_output_test.py
+++ b/integtest/four_process_disabled_output_test.py
@@ -10,8 +10,8 @@ run_duration=20  # seconds
 # Default values for validation parameters
 expected_number_of_data_files=2
 check_for_logfile_errors=True
-min_expected_event_count=run_duration-2
-max_expected_event_count=run_duration+2
+expected_event_count=run_duration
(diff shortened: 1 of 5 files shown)


eflumerf (commented):
Would it be better to define "expected event count" and "tolerance" for the check_event_count check?

philiprodrigues (approved):
Sorry for the delay in reviewing. This looks good to me too, though I agree with @eflumerf that expected/tolerance might be a little nicer than min/max (but either works)

bieryAtFnal:
Eric or Phil, I changed the "event count" check to use 'expected value' and 'tolerance', as you suggested.  Could one of you re-confirm that the changes look OK?  Then we can do the merge.
Thanks,
Kurt

eflumerf (approved):
LGTM

PR #111, opened 2021-11-01
Add ".writing" suffix to raw data filenames during the time that they are being actively written

bieryAtFnal:
As part of this, I added a small "file handle" class to wrap the the HighFive::File instance and automatically rename the raw data files from `xyz.writing` to `xyz` when the file handle is destructed.  This necessitated a few changes in other places...

Thinking about this file handle class a little bit more now (as I write this), it might make sense to move the call to new HighFive::File from the `open_file_if_needed()` method to the HDF5FileHandle constructor, but I'll leave that for another time.

### DIFF
--- a/plugins/HDF5DataStore.hpp
+++ b/plugins/HDF5DataStore.hpp
@@ -91,9 +91,41 @@ ERS_DECLARE_ISSUE_BASE(dfmodules,
 namespace dfmodules {
 
 /**
+ * @brief HDF5FileHandle is a small helper class that takes care
+ * of common actions on HighFive::File instances.
+ */
+class HDF5FileHandle
+{
+public:
+  inline static const std::string s_inprogress_filename_suffix = ".writing";
+
+  explicit HDF5FileHandle(std::unique_ptr<HighFive::File> file_ptr)
+    : m_file_ptr(std::move(file_ptr))
+  {}
+
+  ~HDF5FileHandle()
+  {
+    if (m_file_ptr.get() != nullptr) {
+      m_file_ptr->flush();
+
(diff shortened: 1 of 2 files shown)


plugins/HDF5DataStore.hpp
```diff
@@ -482,25 +524,47 @@ class HDF5DataStore : public DataStore
...
+        }
+      }
+
       // opening file for the first time OR something changed in the name or the way of opening the file
       TLOG_DEBUG(TLVL_BASIC) << get_name() << ": going to open file " << unique_filename << " with open_flags "
                              << std::to_string(open_flags);
       m_basic_name_of_open_file = file_name;
       m_open_flags_of_open_file = open_flags;
-      m_file_ptr.reset(new HighFive::File(unique_filename, open_flags));
+      unique_filename += HDF5FileHandle::s_inprogress_filename_suffix;
```
eflumerf, line 545: Would it be better to have HDF5FileHandle::HDF5FileHandle take a string argument, so this becomes something like:
```CPP
try {
  m_file_handle.reset(new HDF5FileHandle(unique_filename, open_flags));
} catch {
```
and then you could store the original file name in HDF5FileHandle which makes the logic in its destructor much simpler...

bieryAtFnal:
Eric, I've made two changes.  I modified the HDF5FileHandle constructor as you suggested, and I changed the base branch for this PR from _develop_ to _patch/2.8.2_.  If you have a chance, I'd appreciate it if you can take a look.
Thanks

eflumerf (approved):
Looks good to me. Code builds and has correct behavior when running.

PR #113, opened 2021-11-04
HDF5 raw data file changes: better timestamp in filenames, additional Attributes in files

bieryAtFnal:
Added Attributes to HDF5 raw data files for run_number, file_index, and creation_timestamp; changed the timestamp in the raw data filenames to be different for each file index.

These changes were discussed at the v2.8.2-Dev meeting earlier today.

To check that the new HDF5 file Attributes were present, I used `h5dump -A <filename> | head -77`.


eflumerf (approved):
LGTM

PR #114, opened 2021-11-12
Use new BufferedFileWriter interface in the TPSetWriter

floriangroetschla:
The BufferedFileWriter was slightly changed and the TPSetWriter has to be adapted to work. The change is rather minor and makes the code use less calls to the file writer.


bieryAtFnal (approved):
With these changes, dfmodules builds with the latest readout code.

PR #116, opened 2021-11-12
Added an integrationtest check for all expected HDF5 file Attributes

bieryAtFnal:
and added a call to that check in the usual integtest driver files.


eflumerf (approved):
Tested, looks good

PR #117, opened 2021-11-12
Integtest changes to prepare for data with multiple APAs

bieryAtFnal:
A little while ago, I noticed that on the NetworkManager branches, the code produces data with APA numbers different than zero.  So for example, when there are 3 RUs with 5 DLHs each, the data Fragments in the HDF5 file would have five DataSets within each of APA000, APA001, and APA002. 
The existing integtests can not handle this because they used a rather simple model of looking for HDF5 paths that included a specified string (e.g. "TPC/APA000").
The changes on this branch switch to a  model in which we specify the detector group (e.g. "TPC") and the region prefix name (e.g. "APA"), and the "check" code looks for DataSets underneath "TPC/APA\d+".  This will pick up APA000..APA002. 
As part of this, I gave up on testing for specific DataSet names, and my sense was that this was a good thing.  It seems like that might be too much detail in the integtest.
Included in these changes were the appropriate mods to the new run_check.py script.

### DIFF
--- a/integtest/four_process_disabled_output_test.py
+++ b/integtest/four_process_disabled_output_test.py
@@ -12,20 +12,20 @@ expected_number_of_data_files=2
 check_for_logfile_errors=True
 expected_event_count=run_duration
 expected_event_count_tolerance=2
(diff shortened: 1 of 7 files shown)


eflumerf (approved):
LGTM, four_process_quick_test still behaves as expected

PR #118, opened 2021-11-15
Use the new readout packages

jcfreeman2:


jcfreeman2:
I've confirmed that this builds as part of a N21-11-15 work area repo stack which includes `develop` branches of ndreadoutlibs, readoutlibs, fdreadoutlibs and readoutmodules, as well as `readout_breakout` branches of minidaqapp, dqm, flxlibs, dfmodules, and lbrulibs.

PR #119, opened 2021-12-02
integtest updates

bieryAtFnal:
These changes make use of new features in the integrationtest log_file_checks routines, and add some local enhancements as well.

Overall, those changes include:

- additional printout that shows which test variation is being run
- additional printout that shows which logfile has problems (when there are problems)
- the ability to specify "expected" errors that don't cause a particular test to fail (But there are sentences printed out that notify the user that there was one or more ignored problems).

The tests themselves are probably in pretty good shape.  The frustrating thing is that they don't always run cleanly.  I haven't yet been able to determine whether that is because the tests need to be tweaked or the DAQ code needs to be changed.  But, I'm filing this PR anyway.  My feeling is that it will be good to get these test updates into the develop branch of dfmodules as we continue to work on understanding why there are sometimes warning messages in the logs.


eflumerf (approved):
LGTM

PR #120, opened 2021-12-14
Feature/dfo

eflumerf:
Initial implementation of the DFO


mroda88 (approved):
It looks great. Just one thing. With respect to the initial design, we didn't add the specifications of the hooks to extend the algorithm logic, meaning the functions
```C++
void receive_trigger_complete_token(ipm::Receiver::Response message);
bool has_slot() const;
void assign_trigger_decision(std::shared_ptr<AssignedTriggerDecision> assignment);
```
are not virtual. effectively this renders pointless the json metadata in the TRBData. 
I would add the virtual to make the design clear. 

The code itself is working and well tested.

bieryAtFnal (approved):
LGTM

PR #122, opened 2022-01-13
Add unit tests for the DataFlowOrchestrator DAQModule

eflumerf:


mroda88 (commented):
I read part of if and seems good. I'd like to take the time to read the tests properly. 

I have a comment: this might conflict a bit with the correct time accounting of the DFO that I hope to finish by today. Let's keep it in mind when we plan the merges.

mroda88 (approved):
I run the tests and they were ok for me. 
the tests make sense.

PR #124, opened 2022-01-14
Try to clarify when each time counter is being updated.

eflumerf:
Make sure that no time is "missed" by counters when going around while
loops.


mroda88 (approved):
fantastic

PR #125, opened 2022-01-20
Scalable monitoring

mroda88:
This PR redefines the metrics produced by the TRB so that the metrics can be composed to correct summary plots.


eflumerf (approved):
LGTM. Builds and runs without issues.

bieryAtFnal (approved):
LGTM

PR #126, opened 2022-01-20
TRB: reduce latency for td

mroda88:
This is a simple change in the way the TRB goes to sleep when there is nothing to do. Instead of sleeping, the code actually hangs on the extraction of a new TD from the queue. 
This optimises the speed at which new data requests are sent to readout hopefully reducing the dead time. 

I guess this we want to discuss it in a meeting before merging.


mroda88:
I think it's worth attaching this picture. As you can see the change in the code happened more or less in the middle of the time window and you can see that the occupancy of the trigger decision queue (yellow line in the first 3 plots) dropped quite substantially. 

Screenshot from 2022-01-21 12-37-25

eflumerf (approved):
If this is still needed, LGTM. Might need to run dbt-clang-format.sh based on the whitespace changes I saw.

mroda88:
I added the clang formatting

PR #127, opened 2022-01-24
Use queue name, not inst, in FragmentReceiver

philiprodrigues:
This PR changes FragmentReceiver to not hardcode the input queue name, and instead use the mapping provided in the init command. It'll need a corresponding change to the minidaqapp configuration. (The motivation for the change is for the benefit of the new config generation, which picks its own queue instance names)


mroda88 (approved):
LGTM

PR #128, opened 2022-01-31
Added run number to TimeSync messages from FakeDataProd

bieryAtFnal:
Ron noticed that using the FakeDataProd module from the _dfmodules_ repo as part of a _dunedaq-v2.9.0_ system did not result in data being written to disk.

This can be demonstrated with a confgen command like the following:
```
python -m minidaqapp.nanorc.mdapp_multiru_gen --host-ru localhost -o . --use-fake-data-producers mdapp_4proc_fdp
```

I tracked this down to the fact that the FakeDataProd module did not get updated when we added the run number to the TimeSync message (so the FakeHSI module was discarding all of the TimeSync messages from FDP because they had the wrong run number [0]).

The code on this branch adds the run number to the TimeSync messages from FakeDataProd and restores successful data taking with that module.

This branch is based on the HEAD of the _develop_ branch, so we will need to use a software area that uses a recent nightly build to test this code.

### DIFF
--- a/plugins/FakeDataProd.cpp
+++ b/plugins/FakeDataProd.cpp
@@ -33,7 +33,8 @@ enum
 {
   TLVL_ENTER_EXIT_METHODS = 5,
   TLVL_CONFIG = 7,
-  TLVL_WORK_STEPS = 10
+  TLVL_WORK_STEPS = 10,
+  TLVL_TIME_SYNCS = 12
 };
 
 namespace dunedaq {
@@ -50,6 +51,8 @@ FakeDataProd::FakeDataProd(const std::string& name)
   register_command("conf", &FakeDataProd::do_conf);
   register_command("start", &FakeDataProd::do_start);
   register_command("stop", &FakeDataProd::do_stop);
+
+  m_pid_of_current_process = getpid();
 }
 
 void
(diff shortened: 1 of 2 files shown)


ron003 (approved):
I did:
```source /cvmfs/dunedaq.opensciencegrid.org/setup_dunedaq.sh
setup_dbt dunedaq-v2.9.0
dbt-create.sh -n N22-02-01 work_dir && cd $_    # See /cvmfs/dunedaq-development.opensciencegrid.org/nightly/

cd sourcecode
source ~np04daq/bin/web_proxy.sh
git clone https://github.com/DUNE-DAQ/minidaqapp.git -b dunedaq-v2.9.0
git clone https://github.com/DUNE-DAQ/dfmodules
cd dfmodules; git checkout kbiery/FDP_TimeSync_RunNumber; cd ..
source ~np04daq/bin/web_proxy.sh -u
cd ..

dbt-workarea-env
dbt-build.py
```
And ran several tests -- including tests at 10 Hz and with different values for the number of links and the before and after values -- the data sizes/rates changed accordingly

ron003:
--use-fake-data-producers now produce(s) the expected data.

PR #131, opened 2022-02-04
Bugfix: pass queue name, not inst, to queue_inst

philiprodrigues:
Fixes old nanorc config with the changes introduced by #127


philiprodrigues:
> Fixes old nanorc config with the changes introduced by #127

That is, this fix to `FragmentReceiver` causes the old nanorc config to work again (without changes to the config)

PR #134, opened 2022-02-07
Independent DFO

mroda88:
This PR changes the dfmodules to allow the DFO to live in its own application. 
The main consequence for the DFO is the change in the communication protocol between DFO and MLT.


bieryAtFnal (approved):
Looks good based on the recent set of tests.

PR #136, opened 2022-02-08
JCF: let std::vector worry about the memory management of a fake data…

jcfreeman2:
… block


bieryAtFnal (approved):
This change looks good to me.  
I ran before-and-after tests and verified that the memory usage does not grow with the fix.

PR #139, opened 2022-02-18
adding a python script to convert hdf5 file into binary file.

dingp:
This is a simple script to convert HDF5 files into binary files.

The procedure uses the following two steps:

* Step one, traverse through the HDF5 file using `h5py.File.visititems()` to assemble a list containing paths to both trigger record and fragments in the following format:

```python
[
   {
     'TriggerRecordHeader': 'ithTriggerRecordHeaderPath',
     'Fragments': ['jthFragmentsPath_in_ithTriggerRecord,
                             ...]
   },
   ...
]

e.g. 

[
   {
     'TriggerRecordHeader': 'TriggerRecord00000/TriggerRecordHeader',
     'Fragments': ['TriggerRecord00000/FELIX/APA000/Link00',
                   'TriggerRecord00000/FELIX/APA000/Link01']
   },
   {
    'TriggerRecordHeader': 'TriggerRecord00001/TriggerRecordHeader',
     'Fragments': ['TriggerRecord00001/FELIX/APA000/Link00',
                   'TriggerRecord00001/FELIX/APA000/Link01']
   }
]
```

* Step two, follow the order of the assembled list, get byte arrays from HDF5 file, and write them into the output binary file.

No additional information is written to the binary file. Trigger record header itself contains a number of requested components and a list of components, so the total size is calculable. The trigger record header and fragments both have magic words so that verification of data type can be done on the binary blob.  Data is aligned in sequence as:

```
# TRH_X denotes binary blob of a trigger record header
# FG_X denotes binary blob of a fragment

TRH_0 FG_1 FG_2 ... FG_N
TRH_1 FG_1 FG_2 ... FG_M
...
TRH_O FG_1 FG_2 ... FG_P
```

### DIFF
--- a/scripts/hdf5dump/binary_dump.py
+++ b/scripts/hdf5dump/binary_dump.py
@@ -54,7 +54,7 @@ def read_binary_file(fname):
                 nfrag = 0
                 if ntrh !=0 and g_header_type in ['both', 'fragment']:
                     print(80*'-')
-                    print_fragment_header(bytesbuffer)
+                    print_fragment_header(bytesbuffer, g_list_components)
                 bytesbuffer = bytearray()
                 ntrh += 1
                 if ntrh > g_n_request and g_n_request != 0: break
@@ -62,7 +62,7 @@ def read_binary_file(fname):
                 if nfrag == 0 and ntrh != 0 and g_header_type in ['both',
                                                                   'trigger']:
                     print(80*'=')
-                    print_trigger_record_header(bytesbuffer)
+                    print_trigger_record_header(bytesbuffer, g_list_components)
                 if nfrag != 0 and g_header_type in ['both', 'fragment']:
                     print(80*'-')
                     print_fragment_header(bytesbuffer)
--- a/scripts/hdf5dump/binary_dump.py
+++ b/scripts/hdf5dump/binary_dump.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+
+import argparse
+import datetime
+import h5py
+import struct
+from hdf5_dump import print_fragment_header, print_trigger_record_header
+
+CLOCK_SPEED_HZ = 50000000.0
+
+g_n_request = 0
+g_header_type = "both"
+g_list_components = False
+
+
+def get_record_locations_from_binary_file(fname):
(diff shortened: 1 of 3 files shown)


adam-abed-abud:
Hi Pengfei! I like the idea of converting HDF5 files into binary files. I tested the script and it all looks good to me. 
Maybe we should also consider adding a simple parser for the binary files?

dingp:
Hi Adam, thanks for reviewing the script! It was for folks working on FPGA firmwares to process the SNB data. They would like to avoid the overhead of dealing with HDF5.

Having another script parsing/dumping the binary file is a good cross check, and will certainly be useful. I'll add that to the PR soon.

dingp:
@adam-abed-abud I have added a `binary_dump.py` script. It takes similar options as `hdf5_dump.py`, except the `-c/--check-fragments` option. It also takes an additional option, `-d/--debug`, which will print out the location of each header inside the binary file.

The script should give you identical results on the converted binary file as if you run `hdf5_dump.py` on the source HDF5 file.

adam-abed-abud:
Hi Pengfei! Thanks a lot, this feature looks very good. 
I tested both the `hdf5_to_binary.py` and the `binary_dump.py` scripts and I did not encounter any issues. I'm going to merge the pull request.

wesketchum:
I can confirm I also saw no issues when I ran it over TriggerRecord files with the new filelayout structure as, but this is only working for TriggerRecords, not TimeSlices.

Probably in the future, this should be moved to `rawdatautils` and should make use of hdf5libs structures (e.g. HDF5RawDataFile).

dingp:
Thanks @adam-abed-abud and @wesketchum! 

@wesketchum  I guess at some point in the future, we will have python bindings to `hdf5libs`, so python script can be guaranteed to use the latest file structures. But I think maintaining this tool (`hdf5_dump.py` etc) by manually incorporate the new structures is still worth the effort. My understanding is this is a convenient "external" tool which does not depend on the rest of the DAQ release so that anyone can quick peak at the data.

Can you point me to a test HDF5 file with `TimeSlices`? I'll update `hdf5_dump.py, binary_dump.py` and `hdf5_to_binary.py`.

PR #140, opened 2022-02-28
TRB changes for the connection between DQM and DF

jmcarcell:
Goes together with 

- https://github.com/DUNE-DAQ/dqm/pull/81
- https://github.com/DUNE-DAQ/daqconf/pull/3
- https://github.com/DUNE-DAQ/appfwk/pull/215
- https://github.com/DUNE-DAQ/dfmessages/pull/44 (already merged, not in the nightly until tomorrow)

### DIFF
--- a/plugins/TriggerRecordBuilder.hpp
+++ b/plugins/TriggerRecordBuilder.hpp
@@ -19,7 +19,6 @@
 #include "dfmessages/TriggerDecision.hpp"
 #include "dfmessages/Types.hpp"
 #include "dfmessages/TRMonRequest.hpp"
-#include "ipm/Receiver.hpp"
 
 #include "appfwk/DAQModule.hpp"
 #include "appfwk/DAQSink.hpp"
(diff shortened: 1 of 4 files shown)


plugins/TriggerRecordBuilder.hpp
```diff
@@ -230,6 +238,12 @@ class TriggerRecordBuilder : public dunedaq::appfwk::DAQModule
   // Run information
   std::unique_ptr<const daqdataformats::run_number_t> m_run_number = nullptr;
 
+  // Monitoring related variables
+  std::mutex m_mon_mutex;
+  std::string m_mon_connection;
+  //std::map<std::string,dfmessages::TRMonRequest> m_mon_requests;
```
mroda88, line 244: Please get rid of the comment. I would have left it, but since I'm making the comment about the dependency let's fix this as well.

mroda88 (changes requested):
@jmcarcell can you remove the ipm header? It's a bit nasty here and we don't really need it.

mroda88 (approved):
fantastic, thanks!

jmcarcell:
Actually `ipm` is used in the callback that dispatches the requests from DQM so I'm adding it back, saw it right after pushing the previous one but you were too fast :)

mroda88 (approved):
No worries

PR #141, opened 2022-02-28
dfmodules integtest changes for the switch to the daqconf package

bieryAtFnal:
The primary purposes of these changes is to switch the use of the daqconf_multiru_gen script in the daqconf repo.

There are also some reliability and usability improvements in six_process_stop_start_test.py which are unrelated, but useful.


eflumerf (approved):
Tested, working. LGTM

PR #150, opened 2022-03-08
Complete data information from apps

mroda88:
This PR allows the monitoring of the details of each DF applications from the DFO point of view in a dynamic way.


mroda88:
This is just an optimisation related to monitoring, so it's not necessary for 2.10. It's not a bug fix and it's not a required features for 2.10. If necessary it can wait after 2.10

eflumerf (approved):
Works in my testing, though the info_file_collator.py script will need an update as well.

PR #151, opened 2022-03-08
dfmodules integtest changes

bieryAtFnal:
This PR has three changes:

- an initial, simple integtest for long-window-readout
- a fix for occasional problems when comparing the value of the creation_timestamp HDF5 Attribute with the time string in the filename
- an increase in the number of tokens when running with SW TPG in six_process_stop_start_test, to avoid unnecessary warnings about trigger inhibits.

### DIFF
--- a/integtest/long_window_readout_test.py
+++ b/integtest/long_window_readout_test.py
@@ -1,0 +1,79 @@
+import pytest
+import os
+import re
+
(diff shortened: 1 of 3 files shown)


mroda88 (approved):
Both the new `long_readout_window` and the changes in `six_process_stop_start` are working for me. I did the test on np04 machines.

PR #162, opened 2022-03-30
Updated TPStreamWriter to use the new TPSet run_number data member to…

### LINKED ISSUE #160: Discard stale TPSets in TPStreamWriter once TPSet gets run number field
Marco first noticed that TPSets can leak from one run into the next.
A short-term fix was implemented in v2.10.0, but a check on the run number in the TPSet header will help us make a better fix.

bieryAtFnal:
… check for stale TPSets and discard them.

To test my changes, I included the following repos & branches in a software area based on the most recent nightly build:

- dfmodules kbiery/add-tpset-run-number
- fdreadoutlibs philiprodrigues/add-tpset-run-number
- readoutlibs develop
- readoutmodules develop
- trigger philiprodrigues/add-tpset-run-number
- triggeralgs develop

To see the relevant debug messages from TPStreamWriter, I used the following TRACE commands

```
tonM -n TPStreamWriter DEBUG+21
tonM -n TPStreamWriter DEBUG+22
tlvls | grep TPStream  # to confirm that the previous 2 commands worked
```

Of course, we should coordinate with Phil when it comes time to merge this change to _develop_ since it will need to be coordinated with merges in _trigger_ and _fdreadoutlibs_.

### DIFF
--- a/plugins/TPStreamWriter.cpp
+++ b/plugins/TPStreamWriter.cpp
@@ -83,25 +83,6 @@ TPStreamWriter::do_start(const nlohmann::json& payload)
   rcif::cmd::StartParams start_params = payload.get<rcif::cmd::StartParams>();
   m_run_number = start_params.run;
 
-  // 10-Mar-2022, KAB: we have noticed TPSets leaking from one run into the next.
-  // A nice solution would be to check the run number associated with each TPSet
-  // and discard ones from a previous run, but we don't have the run number in
-  // TPSets currently. So, instead, we'll try popping stale TPSets off the queue here.
-  // I don't want this loop to run forever and hang the program if something
-  // changes in the order of App Start commands, so I'll limit its duration to
-  // a second or two...
-  auto start_time = std::chrono::steady_clock::now();
-  auto now = std::chrono::steady_clock::now();
-  trigger::TPSet tpset;
-  while (m_tpset_source->can_pop() && (now - start_time) < std::chrono::seconds(2)) {
-    now = std::chrono::steady_clock::now();
-    try {
-      m_tpset_source->pop(tpset, m_queue_timeout);
-    } catch (appfwk::QueueTimeoutExpired&) {
-      break;
-    }
-  }
-
   // 06-Mar-2022, KAB: added this call to allow DataStore to prepare for the run.
   // I've put this call fairly early in this method because it could throw an
   // exception and abort the run start.  And, it seems sensible to avoid starting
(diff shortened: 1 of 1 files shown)


bieryAtFnal:
I should have mentioned that we need to have multiple runs within a single DAQ session to see the problem.

Here are some sample commands to help with that:
```
daqconf_multiru_gen --host-ru localhost --host-ru localhost --host-df localhost --host-df localhost -d $PWD/frames.bin -o . -s 10 -n 2 --enable-software-tpg --enable-tpset-writing mdapp_2x2_swtpg_tpsw

runbase=100; runduration=20; waitAfterStop=2; let R1=${runbase}+1; let R2=${runbase}+2; let R3=${runbase}+3; let R4=${runbase}+4; nanorc mdapp_2x2_swtpg_tpsw boot init conf start ${R1} wait ${runduration} stop wait ${waitAfterStop} start ${R2} wait ${runduration} stop wait ${waitAfterStop} start ${R3} wait ${runduration} stop wait ${waitAfterStop} start ${R4} wait ${runduration} stop wait ${waitAfterStop} scrap terminate
```

mroda88 (approved):
I tested interactively and with the following command sequences:
```
boot init conf start 500 wait 1000 stop start 501 wait 100 stop wait 1 start 502 wait 100 stop terminate
``` 
```
boot init conf start 600 wait 100 stop wait 100 start 601 wait 100 stop wait 100 start 602 wait 100 stop terminate 
```
I didn't see any error or warning and all seems to work. 

I noticed something that I think it's worth sharing. In between a `stop` and a new `start` the queue of tpsets in the tpwriter app saturates completely
Screenshot from 2022-04-01 12-49-35

Is it something we are worried about? 
What worries me the most is that since the queue completely saturates, something should fail to push, but no warnings appear. 
If we are happy with this behaviour I don't see any problem at all.

eflumerf (approved):
LGTM. TRACEs show appropriate run numbers and discarding when not matching.

bieryAtFnal:
To respond to Marco's concern about the number of TPSets left in the tpwriter_tpsets_into_writer_0output_to_tpswritertpset_source queue between runs...

I looked at the configured size of this queue, and it seems to be 1000.  So, my sense is that what we're seeing in the plots that Marco attached is that there are ~100 TPSet messages that arrive at the tpwriter app between the time that it receives the Stop command and the time that the Readout apps receive the Stop command.  Since 100 < 1000, we don't see any warning messages about the queue filling up.  

Of course, we can certainly imagine situations in which we *could* fill up that queue, but that situation seems to me to be part of the whole we-should-improve-the-handling-of-Stop-conditions discussion that we will have some day.  And, as such, my sense is to go ahead with the merge of the code changes covered by this PR.

mroda88:
I agree that the problem is related to transition. Just a comment that plot is the relative occupancy, so we do have 1000 (or more) TPSets in those queues. But it's not that important I think.

PR #165, opened 2022-04-07
TPStream monitoring

### LINKED ISSUE #163: Missing opmon metrics for tpwriter
At the moment there are not metrics to monitor the flow of TPs. 
I think the most urgent thing is at least track how much we are writing on disk. 

I can take care of it and do the integration with the dashboards, if no one has time.

mroda88:
this adds the monitoring for the TP stream writer.

### DIFF
--- a/plugins/TPStreamWriter.cpp
+++ b/plugins/TPStreamWriter.cpp
@@ -10,6 +10,7 @@
 #include "dfmodules/CommonIssues.hpp"
 #include "dfmodules/TPBundleHandler.hpp"
 #include "dfmodules/tpstreamwriter/Nljs.hpp"
+#include "dfmodules/tpstreamwriterinfo/InfoNljs.hpp"
 
 #include "appfwk/DAQModuleHelper.hpp"
 #include "daqdataformats/Fragment.hpp"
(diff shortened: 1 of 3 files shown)


bieryAtFnal:
With these code changes, I confirmed that tpwriter metrics are written to disk, when opmon is configured to do that.

mroda88:
Here is how the final panel looks like
Screenshot from 2022-04-08 13-36-41

bieryAtFnal:
Marco, thanks for the plot.  I was able to confirm that metrics showed up in the Grafana Overview dashboard from my partition at NP04 for the tpwriter after you updated the dashboard.
One question about the values shown in the Grafana plot, though.  I expected that the bytes output rate from the tpwriter would be larger than from the data writers in a basic test.
I confirmed this with opmon results written to my local disk.  This seems to indicate that it is something with the display, rather than something in the reporting of the bytes_output.  
In my test, the TRB output_bytes per 10-second sample was around 2-5 MB.  For the TPStreamWriter, it was 21-30 MB.  But, the Grafana plot showed the tpwriter values being less than the TRB.  
Thoughts?

mroda88:
You are absolutely right, Apologies. In restoring the dashboard while listening to the meeting I was not paying too much attention and the math behind the plot was done wrong. 
I fixed it now:
Screenshot from 2022-04-08 22-03-06
The problem was just in the dashboard.

roland-sipos (approved):
LGTM

PR #166, opened 2022-04-13
fix unittest for custom commands

glehmannmiotto:
to be merged after appfwk glm/custom_commands


eflumerf (approved):
LGTM. Since this is a multi-repo changeset, John or Pengfei should be contacted to manage the merges.

PR #167, opened 2022-04-14
DFO Roundrobin

### LINKED ISSUE #159: Round-robin distribution of events to TRBuilders
It would be nice to have the option to specify a round-robin distribution of TriggerRecords to TRBuilders by the DFO (and maybe this should be the default).

This would avoid the situation in which certain TRBs never receive events (at lower trigger rates).  And, it might produce plots in Grafana that are more in line with what viewers might expect.

It would certainly help with integration tests, in that similar numbers of TRs would be expected in each TRB.

mroda88:
this solves #159 
I tested running two cases:
 - single DF application with standard 1Hz rate
 - 6 DF applications with with 1Hz rate
 
 All the applications were used successfully. 
 
 I implemented the logic so that the cycle start from scratch on each stop -> start transition. 

I attach the plots to show that applications were used with the same rate. 
Screenshot from 2022-04-14 11-54-23
Screenshot from 2022-04-14 11-53-51

### DIFF
--- a/plugins/DataFlowOrchestrator.cpp
+++ b/plugins/DataFlowOrchestrator.cpp
@@ -206,29 +206,31 @@ DataFlowOrchestrator::find_slot(dfmessages::TriggerDecision decision)
   // this find_slot assings the decision with a round-robin logic
   // across all the available applications.
   // Applications in error are skipped.
-
+  // we only probe the applications once.
+  // if they are all unavailable we retur to the caller 
+  // before looping again
+  
   std::shared_ptr<AssignedTriggerDecision> output = nullptr;
+  unsigned int counter = 0 ;
+ 
+  auto candidate_it = m_last_assignement_it;
+  if (candidate_it == m_dataflow_availability.end())
+    candidate_it = m_dataflow_availability.begin();
 
-  auto last_it = m_last_assignement_it;
-  if (last_it == m_dataflow_availability.end())
-    last_it = m_dataflow_availability.begin();
+  while ( output == nullptr && counter < m_dataflow_availability.size() ) {
 
-  auto candidate_it = last_it;
-
-  do {
-
+    ++counter;
     ++candidate_it;
     if (candidate_it == m_dataflow_availability.end())
       candidate_it = m_dataflow_availability.begin();
-
+    
     if (candidate_it->second.is_busy())
       continue;
-
+    
     output = candidate_it->second.make_assignment(decision);
     m_last_assignement_it = candidate_it;
-    break;
-
-  } while (candidate_it != last_it);
+    
+  }
 
   return output;
 }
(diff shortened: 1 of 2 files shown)


plugins/DataFlowOrchestrator.cpp
```diff
@@ -202,31 +203,32 @@ std::shared_ptr<AssignedTriggerDecision>
...
-  }
+  auto last_it = m_last_assignement_it;
+  if (last_it == m_dataflow_availability.end())
+    last_it = m_dataflow_availability.begin();
 
-  if (candidate != m_dataflow_availability.end()) {
-    output = candidate->second.make_assignment(decision);
-    m_last_sent_td_connection = candidate->first;
-  }
+  auto candidate_it = last_it;
```
eflumerf: I think we can get rid of `last_it`. Have this line be `auto candidate_it = m_last_assignment_it;`, and the while loop predicated on `while(candidate_it != m_last_assignment_it)` or better, instead of a do..while, use `while(output == nullptr)` to avoid confusing constructs.

plugins/DataFlowOrchestrator.cpp
```diff
@@ -202,31 +203,32 @@ std::shared_ptr<AssignedTriggerDecision>
...
+
+  do {
+
+    ++candidate_it;
+    if (candidate_it == m_dataflow_availability.end())
+      candidate_it = m_dataflow_availability.begin();
+
+    if (candidate_it->second.is_in_error())
+      continue;
```
eflumerf, line 228: Should we also check if the TRB is over a busy threshold here?

plugins/DataFlowOrchestrator.cpp
eflumerf: To restrict the loop in the case where all TRBs are full or in error, maybe have it be `while(output == nullptr && counter < m_dataflow_availability.size())`

mroda88: So the `last_it` here has an important role. Since `m_last_assignment_it` can point to a non-valid entry of the map (the end) when it is the beginning of the run. 
`last_it` has the benefit of always being an iterator pointing to a valid entry of the map. 

I prefer to avoid entering a possible infinite loop, so the while( output == nullptr) is not ok with the rest of the logic. 
Here we just try to find a suitable candidate. If all the applications are in error or busy (see next comment) the search simply fails and it's up to higher level code to decide what to do.

plugins/DataFlowOrchestrator.cpp
mroda88: It might change slightly deviate the logic of the round-robin when the applications are all busy. Let me try and see if I see an acceptable behaviour. 
I'll come back here when I have done some tests.

mroda88: Ok, I added the change. As at high rate, all the applications are used anyway and the issue brought up is not valid anyway.

bieryAtFnal (approved):
I confirm that the number of events received by each of the three DF Apps in an integration test are now very close to each other.

mroda88:
Ok, Changes done. @eflumerf have a look if the code is more clear now.

eflumerf (approved):
LGTM

PR #168, opened 2022-04-15
Renaming and updating of integration tests in dfmodules

### LINKED ISSUE #161: Rename integtests and create additional ones
details to be provided...

bieryAtFnal:
As requested in Issue #161 , this issue provides (hopefully) better names for several integtests.
In addition, a new test was added for multiple DF apps.  (3ru_3df; this is different than the 3ru_1df test, which is more focused on RU behavior).
And, there is another attempt to fix the bug in the validation of the creation_timestamp attribute in the HDF5 files.


eflumerf (approved):
LGTM

PR #169, opened 2022-04-19
Tweaked multi_output_file_test parameters to avoid warning messages.

bieryAtFnal:
Tested on mu2edaq14, mu2edaq11, iceberg01, and np04-srv-013.


eflumerf (approved):
LGTM
This test now runs on my development node as well.

PR #173, opened 2022-05-17
Fix connection with DQM

jmcarcell:
This line was deleted with the changes to iomanager. Goes together with changes in `daqconf` that aren't ready yet


jmcarcell:
Not needed

Issue #177, opened 2022-05-22
Lost TriggerRecord in 3ru_1df test with N22-05-22 nightly?


bieryAtFnal:
When I run the following test with the N22-05-022 Spack nightly build, 

```
daqconf_multiru_gen -d ./frames.bin -o . -s 10 -n 3 -b 1000 -a 1000 --latency-buffer-size 200000 --host-ru localhost --host-ru localhost --host-ru localhost --region-id 0 --region-id 1 --region-id 2 mdapp_3ru_1df
curl -o frames.bin -O https://cernbox.cern.ch/index.php/s/7qNnuxD8igDOVJT/download
nanorc mdapp_3ru_1df/ boot ${USER}_test init conf start 201 wait 20 stop wait 3 scrap terminate
```

I see the following error in the DF App log:

```
log_dataflow0_3339.txt:2022-May-22 07:10:53,110 ERROR [bool dunedaq::dfmodules::TriggerRecordBuilder::send_trigger_record(const dunedaq::dfmodules::TriggerId&, std::atomic<bool>&) at /home/biery/dunedaq/22MayDevTest/sourcecode/dfmodules/plugins/TriggerRecordBuilder.cpp:685] trigger ID 20-0/201 could not be sent to writing and it's lost
```

philiprodrigues:
I'm able to reproduce this issue. There are at least two problems, if I'm understanding correctly. First, `TriggerRecordBuilder` has this:

https://github.com/DUNE-DAQ/dfmodules/blob/cc05877fd75d5d4a4e493529b83b1750ed13c2b3/plugins/TriggerRecordBuilder.cpp#L670-L680

which won't even attempt the send if it's called after the `running` flag is set to false. I assume a `do {...} while (...);` loop is intended.

The problem upstream of that is that the TC for that TR is not being returned, which must be some problem in `trigger` or `readoutlibs`. I recently wrote code to handle exactly this case, so I'm not sure what's up. Investigating.

mroda88:
@philiprodrigues you are correct, that should definitely be a `do while` loop. I can fix it quickly if you want to work on the readoutlib side.

philiprodrigues:
Two other problems contributed to this issue:
1. I made a mistake in handling waiting data requests at run stop in `readoutlibs`. This is fixed by https://github.com/DUNE-DAQ/readoutlibs/pull/28
[...]

PR #178, opened 2022-05-23
Correct sending loop

mroda88:
this should fix issues #177


bieryAtFnal:
As I reported in Issue #179, I'm seeing a warning message like the following:
```
Problem(s) found in logfile /tmp/pytest-of-biery/pytest-680/run1/log_dfo_3335.txt:
2022-May-23 20:54:26,813 WARNING [virtual void dunedaq::dfmodules::DataFlowOrchestrator::receive_trigger_complete_token(const dunedaq::dfmessages::TriggerDecisionToken&) at /home/biery/dunedaq/23MayDevTest/sourcecode/dfmodules/plugins/DataFlowOrchestrator.cpp:266] DataFlowOrchestrator encountered run number mismatch: recvd (101) != 102 from trigger_decision_0
```
in the 3ru_1df_multirun_test....  Not sure if that is unrelated to this change yet...  I saw this error both with, and without, the changes that are part of this PR...

bieryAtFnal (approved):
I haven't seen any problems related to this change.

PR #184, opened 2022-06-01
Fixed bug/copy-paste-error in assignment of expected_fragment_count in 3ru_1df test

bieryAtFnal:
Found and fixed this typo from when I converted the 3ru_1df test to support TA and TC fragments.


eflumerf (approved):
LGTM, test now has different failures

PR #187, opened 2022-06-02
using try_receive instead of receive for TRB

### LINKED ISSUE #186: TRB very inefficient with exceptions
The loop in the TRB is controlled by a number of receive calls that have 0 timeout. 
Due to what has been discussed yesterday, this is far from optimal. 
Since we now have dedicated methods to solve this, we should use them

mroda88:
this solves issue #186

### DIFF
--- a/plugins/TriggerRecordBuilder.cpp
+++ b/plugins/TriggerRecordBuilder.cpp
@@ -370,17 +370,20 @@ TriggerRecordBuilder::read_fragments()
 
   for (unsigned int j = 0; j < m_fragment_inputs.size(); ++j) {
     
-    std::unique_ptr<daqdataformats::Fragment> temp_fragment;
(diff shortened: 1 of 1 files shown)


plugins/TriggerRecordBuilder.cpp
```diff
@@ -370,17 +370,20 @@ TriggerRecordBuilder::read_fragments()
...
+      ers::error(e);
       continue ;
     }
 
+    if ( ! temp_fragment ) continue ;
+    
     new_fragments = true;
     
-    TriggerId temp_id(*temp_fragment);
+    TriggerId temp_id(**temp_fragment);
```
eflumerf: Use .value() to dereference std::optional

eflumerf (approved):
LGTM, builds and doesn't make the integration tests *worse*

PR #189, opened 2022-06-06
Added Tardy input set message to Trigger App exclusions in 3ru* integtests

bieryAtFnal:


philiprodrigues (approved):
The code looks correct to me, and in a test of `3ru_1df_multirun_test.py`, I see the tardy warnings in the log, but the integration test passes, as intended

PR #196, opened 2022-06-21
removing tardy tazipper exclusion from certain integration tests

bieryAtFnal:
One of the work-arounds that we implemented in v3.0.0 was to ignore "tardy tazipper" warning messages in dfmodules/integtest log files.

Phil has since fixed the tardy tazipper handling, and the warning messages no longer appear.  This PR simply backs them out of the relevant dfmodules/integtest tests.

As an aside, it also adds some code to the multi_output_file_test to clean up the raw data HDF5 files at the end of the integration test, so that we don't fill up /tmp with such files.

### DIFF
--- a/integtest/3ru_1df_multirun_test.py
+++ b/integtest/3ru_1df_multirun_test.py
@@ -34,8 +34,7 @@ triggertp_frag_params={"fragment_type_description": "Trigger with TPs",
                        "min_size_bytes": 80, "max_size_bytes": 16000}
 ignored_logfile_problems={"dqm": ["client will not be able to connect to Kafka cluster",
                                   "Unexpected Trigger Decision", "Unexpected Fragment"],
-                          "trigger": ["zipped_tpset_q: Unable to push within timeout period",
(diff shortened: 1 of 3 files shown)


jcfreeman2 (approved):
The removal of the tardy tazipper warning from the exclusion list passes a simple eyeball test, and files are indeed listed before deletion, e.g.
```
..F============================================
Listing the hdf5 files before deleting them:
============================================
Filesystem      Size  Used Avail Use% Mounted on
/dev/sdb4        20G  4.0G   16G  21% /tmp
--------------------
-rw-r--r-- 1 jcfree mu2e 206845040 Jun 21 16:00 /tmp/pytest-of-jcfree/pytest-1/run0/swtest_run000101_0000_dataflow0_20220621T210031.hdf5
-rw-r--r-- 1 jcfree mu2e 139398944 Jun 21 16:01 /tmp/pytest-of-jcfree/pytest-1/run0/swtest_run000102_0000_dataflow0_20220621T210117.hdf5
--------------------
```

PR #22, opened 2021-02-02
Eflumerf/dataformats dfmessages styleguide compatibility

eflumerf:
Resolves #19

PR #197, opened 2022-06-22
Added a check in the DataWriter code that TriggerRecords have the expected run number

bieryAtFnal:
also ran clang-format on the header; also added run number to several messages.

These changes were helpful in testing iomanager PR #22.


mroda88 (approved):
The code looks good to me. 
I tested it and it worked like a charm.

PR #201, opened 2022-06-23
DFO stop transition

### LINKED ISSUE #198: DFO should not complete its stop transition until all TriggerCompletes are received
The DFO should refuse to stop until all of its bookkeeping has been completed. This may cause the CCM processes to declare a failed transition if timeouts occur.

mroda88:


mroda88:
According to yesterday's meeting, this PR provides a stop transition for the DFO that waits until all the tokens are delivered before returning. 
For safety reason, a timeout has been added and proper errors will be flagged when things are not completed.

Also according to the discussion, same warnings were raised to errors. 

Only preliminary tests has been done at the moment, when all the tests are done, I'll convert the draft into a proper PR

eflumerf (approved):
I ran the 3ru_3df_multirun_test with my changes in https://github.com/DUNE-DAQ/daqconf/pull/118, and all tests were successful

mroda88:
Thanks @eflumerf ! It's very good to know. Before preparing the proper PR, I will need to add integration tests for the new features of the DFO. Also I think I will want to do some large scale tests too to check that things are stable.

bieryAtFnal:
For reference, I've been able to demonstrate the problem with the following confgen command:
```
daqconf_multiru_gen -t 5.0 --use-fake-data-producers -o <location on a slow disk> -s 10 -n 5 -b 100000 -a 100000 mdapp_fdp
```
where the 'slow disk' location is the key.  We want the writing of the data to be slower than its creation.

and the following nanorc command
```
runbase=100; runduration=35; waitAfterStop=2; let R1=${runbase}+1; let R2=${runbase}+2; nanorc mdapp_fdp boot ${USER}-test init conf start ${R1} wait ${runduration} stop wait ${waitAfterStop} start ${R2} wait ${runduration} stop wait ${waitAfterStop} scrap terminate
```

For me, `egrep 'ERROR|WARNING' log_*` shows run-number-mismatch warnings from the DFO and trigger-inhibit messages from the trigger.

mroda88:
I'm going to write here a note about the stop timeout I introduced. Using the instructions from Kurt, I tested some values of that timeout. 
With 100 ms, the system does not always collect the TD in time and we get the newly added errors that the corresponding TD are not completed. 
With 1 s, the system finds them all, but clearly this is test specific. It might be that on a slower or more complex system, 1 s is not enough. 

Shall we say we set the default to 10s? Too much? I'm happy to take feedback.

mroda88:
I performed also the large scale system test to see if this helped to keep the system ok. 
Unfortunately the run generated incomplete trigger records in the middle of the run, so with the new changes we see errors at stop enumerating what is not complete. 
When we have the new order we might be able to understand more. 
But I think for now the PR is ready, at least from my side.

bieryAtFnal:
I've run tests of these changes in conjunction with the updated Stop order in `daqconf` ([daqconf PR 118](https://github.com/DUNE-DAQ/daqconf/pull/118))

A test of the confgen and nanorc commands that I listed above looked good, in terms of expected number of TriggerRecords and expected errors/warnings in the logs.  I'd like to run this test again with various TRACE messages enabled and double-check that the underlying behavior looks good; I'll try to do that tomorrow.

In addition, I went back to the modified long_window_readout integtest that I was working on when I originally noticed this problem (with the output location being a slow, NFS-mounted disk), and these changes looked very good there.  I did need to locally modify the value of the stop timeout that Marco mentioned above (10->100 seconds), and it might be good to touch base on the preferred timeout value when we next meet.

Lastly, I'd like to take a look at how the MLT-DFO interaction works at Start time with the new Start order, mainly out of curiosity.  E.g., does the MLT start in an inhibited state and then get tokens from the DFO when the DFO starts, or what? (new start order is trigger, hsi, readout, dataflow, dfo).

bieryAtFnal:
After further tests, things still look good.  

A few of my test runs had occurrences of the waiting-for-trigger-complete-message timeout, and in those cases, the TriggerDecisionToken (trigger-complete) messages from the late TriggerRecords arrived in the next run.  The messages for those were a little confusing, so I added a little text to the error message.  For better or worse, I took the opportunity to add in several additional debug messages that I've been using in debugging problems like these.  Those are on a new branch, and I will file a PR for them.

mroda88:
Hi Kurt, thanks! I wanted to check what happens on the TRB side today, but I got stuck with other stuff. I'll try to do it tomorrow before the meeting. I had a look at your PR, I think it's ok, although I didn't test it. But for sake of completeness I think I will merge it now.

bieryAtFnal (approved):
My tests with these changes look quite good.
As we've discussed, there are a couple of follow-on changes that we will work on separately.

PR #202, opened 2022-06-28
Added messages in the DFO to help with debugging

bieryAtFnal:
The only change that is not just a message is the addition of the 500 usec sleep when a slot can not be found.  This is to avoid a tight loop in that case.


mroda88 (approved):
Looks great!

PR #205, opened 2022-07-07
More monitoring and avoid null assignment

mroda88:


eflumerf (approved):
Unit tests and a quick nanorc run look good

bieryAtFnal (approved):
LGTM too.
I ran tests in which I buggered the ModuleLevelTrigger to ignore the DFO Busy message and saw that the DFO continued to accept triggers (with a system configuration that would have otherwise produced Inhibits).
```
daqconf_multiru_gen -t 5.0 --use-fake-data-producers -o /data1/kurt -s 10 -n 5 -b 100000 -a 100000 mdapp_fdp_stressDFO
```
(/data1/kurt is a slow, NFS-mounted disk area)

PR #211, opened 2022-07-19
Mroda/new bytes

mroda88:
This will sync the branch with develop and add a new tag on CMakeList. 
According to what john told me.


eflumerf (approved):
LGTM. Updating prep-release branch to previous tag and adding Marco's change.

PR #213, opened 2022-07-22
Update command order in tpstream_writing_test. Add test method which

eflumerf:
gathers and tests tpstream output files.


bieryAtFnal (approved):
LGTM

PR #216, opened 2022-08-17
Update counts of trigger fragments: there are now always exactly 3

philiprodrigues:
Previously, there was one TP/TA trigger fragment for each producer of
those. Now, all of the TPs go into one fragment and all of the TAs go
in another. Along with the TC fragment, that makes 3 trigger fragments
per event


bieryAtFnal:
Phil, your changes look good.  (though, I did need to merge recent changes from the develop branch into this branch)
However, I'm having trouble getting consistently successful runs of the tests, especially those with SWTPG.  One of the problems is likely due to the (relatively new) merging of triggers in the MLT.  Sometimes the integtest fails because the WIB fragments are too large.  I suspect that a periodic HSI trigger is being merged with a TP-based trigger, and that results in a wider readout window than the integtest expects.  I'm working on updating the allowed range of WIB fragment sizes...

bieryAtFnal (approved):
Tests of these changes look reasonable.

PR #217, opened 2022-08-26
put application name and module name in filename to support multiple …

glehmannmiotto:
…data writers per dataflow application; changed from callback to receive API to use the properties of a MPMC queue between TRB and data writer modules


ron003:
I've started looking at this.

ron003 (approved):
ran additional tests -- all seems to work as expected.

ron003:
with changes to hdf5libs --- now needed also

PR #219, opened 2022-09-01
Update expected number of TP fragments

philiprodrigues:
Was previously expecting

(number_of_data_producers*number_of_readout_apps)+number_of_readout_apps+1,

which is one for each TP link, one each for TAs and one for TCs. But
the actual check only looks for TP fragments, so we just expect

(number_of_data_producers*number_of_readout_apps)


eflumerf:
Should we add explicit checks for TA and TC Fragments? (Maybe as another PR)

eflumerf (approved):
LTGM

PR #220, opened 2022-09-01
Set the det_crate value in hw map to the readout app #

philiprodrigues:
Previously, all links in the hardware map had det_crate set to
zero. The daqconf uses det_crate to decide which links to aggregate
into a TA maker in the trigger. When all the det_crate values are the
same, all of the TP links are aggregated into one TA  maker. This is
fine - it's a perfectly legal configuration - but doesn't exercise the
functionality of having multiple TA makers, which is a more typical
configuration.

I've set the crate value to be the same as the app number, but I see
that the card number is also set to the same value, so I'm not sure
whether this is the right thing to do.


eflumerf (approved):
Since this is a completely-fake configuration, I think that it's perfectly valid to do it this way.
LGTM

PR #227, opened 2022-09-19
Integtest fixes, mainly in the multi_output_file_test

bieryAtFnal:
This PR has several changes

Several fragment size ranges were updated in the multi_output_file_test to take into account the larger readout window that is used in this test (compared to 3ru_3df*test, for example).

The calculation of the configured number of tokens was fixed in three integtests.  That seems to have become broken in the conversion to daqconf.json.

I added debug printout of the used_slots (count) in DataFlowOrchestrator.cpp.


eflumerf (approved):
LGTM. Kurt explained to me that the reason for the token scaling is because all of the links are reading the same frames.bin file, so in SWTPG config, they'll all generate TCs from all links at the same time, slamming the dataflow system.

PR #232, opened 2022-09-28
Always set the FragmentType to kSW_TriggerPrimitive in the TPStreamWriter

### LINKED ISSUE #231: TPwriter mislabeling time slices as FW_TriggerPrimitives
The TPWriter is currently setting the timeslice type according to whether fw or sow TPs are used, but actually the type is SW_TriggerPrimitives in either case.

bieryAtFnal:
This branch has the minimal fix for Issue 231.

We can make more thorough changes to remove knowledge of the TP source from the TPStreamWriter code now (as part of v3.2.0) or later.  Let me know if we want that to be done as part of v3.2.0.  It will affect dfmodules and daqconf.

I haven't yet updated the "base" branch for this PR to be the v3.2.0 release-prep branch, but I can create the release-prep branch and update this PR, if that is desired.

### DIFF
--- a/integtest/fake_data_producer_test.py
+++ b/integtest/fake_data_producer_test.py
@@ -20,7 +20,7 @@ check_for_logfile_errors=True
 expected_event_count=run_duration
 expected_event_count_tolerance=2
 wib1_frag_hsi_trig_params={"fragment_type_description": "WIB", 
-                           "fragment_type": "Unknown",
+                           "fragment_type": "ProtoWIB",
                            "hdf5_source_subsystem": "Detector_Readout",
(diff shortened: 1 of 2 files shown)


bieryAtFnal:
Oops, I let an unrelated change slip in.  I'll fix that in a little while.

bieryAtFnal:
An update:  I've created the prep-release/dunedaq-v3.2.0 branch and set that as the target of this PR.  And, I've removed the change to fake_data_producer_test.py that accidentally slipped onto this branch.

eflumerf (approved):
LGTM

PR #237, opened 2022-10-03
Removed knowledge of FW TPs in the TPStreamWriter

bieryAtFnal:
since it should only write 'SW' TPs.

This is a more thorough fix for Issue #231.  It is for consideration **after** v3.2.0.  It will be coupled with a PR in the `daqconf` repo.

I created this branch from the v2.9.1 tag of the `dfmodules` repo so that I could get the advantage of the fixes that have been made in testing of the v3.2.0 release.  So, in comparisons with the `develop` branch, we will see the TPStreamWriter changes plus a couple more.


eflumerf (approved):
builds, minimal_system_quick_test and tpstream_writing_test both pass all cases.

PR #243, opened 2022-11-16
Draft of writing time metric

mroda88:


eflumerf (approved):
Works in my test area, don't know if we'll have time to test on a detector before the next temporary freeze.

wesketchum (approved):
Looks good. Pending test at EHN1.

PR #257, opened 2023-01-23
Update usages in unit test to account for new parameter to IOManager::configure

eflumerf:
Follows https://github.com/DUNE-DAQ/iomanager/pull/35


ron003 (approved):
tested at np04

PR #260, opened 2023-02-02
integtest/README frames.bin fix

bieryAtFnal:
Julia noticed that the instructions for fetching the frames.bin file were out of date.  This change fixes that.


jtenavidal (approved):
It is very clear and the new file can be used to run the integration tests.

PR #271, opened 2023-03-03
Modified the TRB code to save the reply conection name at INIT time

bieryAtFnal:
…and removed the reply_connection_name from the TRB config schema.

This is part of the easy fixes that we are doing as part of the ConnSvc consolidation.

This PR should be tested and review in conjunction with one in the daqconf repo.


roland-sipos (approved):
LGTM

eflumerf (approved):
LGTM. Integration tests appear to be OK

PR #280, opened 2023-04-02
fixed the assignment of the token count in several integration tests

bieryAtFnal:
I recently noticed that the multi_output_file_test occasionally generates "trigger inhibit" warnings when running the SWTPG part of its testing.  As part of investigating this, I noticed that the token count that is being requested in the integtest script was not being taken into account.  The problem was that the location of the token_count parameter in the dataflow daqconf schema tree changed, but the integtests were not correspondingly updated  I fixed that problem and tweaked a couple of the token_count calculations.  With these changes, the multi_output_file_test integtest runs more reliably without trigger inhibit warnings.  To test these changes, I ran each of the 5 modified integtests from the develop branch and then from the kbiery/integtest_token_count_fix branch and compared the generated configurations under /tmp/pytest_of_<user>/pytest_N/jsonX and /tmp/pytest_of_<user>/pytest_N-1/jsonX to confirm that values for the DFO busy and free parameters were different from the default values (which were being used before the changes).

### DIFF
--- a/integtest/3ru_1df_multirun_test.py
+++ b/integtest/3ru_1df_multirun_test.py
@@ -93,7 +93,7 @@ conf_dict["readout"]["default_data_file"] = "asset://?label=ProtoWIB&subsystem=r
 
 swtpg_conf = copy.deepcopy(conf_dict)
 swtpg_conf["readout"]["enable_software_tpg"] = True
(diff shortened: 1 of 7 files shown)


plasorak:
That's weird, daqconf shouldn't in principle let you do that, the configuration generation should fail if the key doesn't exist in the schema

eflumerf (approved):
lgtm, tests behave normally

PR #283, opened 2023-04-05
Extend the stop wait time for disabled_output_test.

eflumerf:


bieryAtFnal:
I haven't been able to reproduce the problem without this change, so I haven't been able to run before and after tests yet.  I'll keep trying.

bieryAtFnal:
I tried the unmodified disabled_output_test on mu2edaq13, mu2edaq11, and np04-srv-013, and have only see some very occasional 'Tardy TPSet' warning messages.  (I presume that we're looking for TriggerApp Stop timeouts and subsequent failures, and those I haven't seen.)

A quick side note is that this code change is reducing the time between the following transitions, in a couple of cases:
 - 'disable_triggers'
 - 'drain_dataflow'
 - 'stop_trigger_sources'
 - 'stop'
That is probably understood.  And, of course, the code change *is* increasing the quiet time between the full stop of one run and the beginning of the start transition for the next one.

eflumerf:
My testing today with the unmodified test was all working, so I'm fine with simply closing this.

PR #284, opened 2023-04-06
Add multiple output directory configurations to multi_output_file_tes…

### LINKED ISSUE #245: Add integration test for multiple output directories for a DF app
https://github.com/DUNE-DAQ/hdf5libs/issues/45 indicates that we need to test this

eflumerf:
…t. Resolves #245

I was unable to get dunedaq-v3.2.0 working to do a differential test (i.e. that the added tests fail when using hdf5 without threading support).


bieryAtFnal (approved):
LGTM.
I saw a few failures of the WIBEth_System (Multiple Outputs) test in which there were 5 raw data files instead of 4.  This seemed to be caused by one of the DataWriters getting a larger fraction of the trigger records and crossing the max_file_size threshold of about 1 GB.  Increasing the max_file_size value to 4 GB helped prevent multiple files when one DataWriter gets more of the trigger records.
Another change was to make the cleanup of the raw data files more robust.

PR #285, opened 2023-04-10
Draft integtests that may be useful for testing dunedaq-v4.0.0

bieryAtFnal:
…but are primarily targeted for v4.1.0.

My current thought is that we should keep the existing dfmodules/integtests unchanged at this point in time, since we are already in the testing phase of the v4.0.0 software release.  However, there are tests that have been shown to be useful *and* are in line with changes that we want to make early in the v4.1.0 development cycle, so my idea is to contribute them to a "drafts" subdirectory under the "dfmodules/integtest" directory.

The new version of the minimal_system_quick_test always uses the Connectivity Service.  This test can already be run successfully.

The new version of the 3ru_3df_multirun_test uses DUNE-WIB emulated data instead of ProtoWIB data, always uses the ConnectivityService, and selects a specific DuneWIB frames.bin file using its checksum so that we can reliably know the number of expected TriggerRecords when SWTPG is enabled and the software_tpg_threshold is set to 150.  This test currently fails because of warning messages in the logfiles during the SWTPG part of the test.  The fix for those warnings is a one-line change in the fdreadoutlibs repo, so we hope that we can have this new integtest running successfully soon.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,4 @@ Structs.hpp
 # integtest temporary files
 frames.bin
 wib2-frames.bin
-integtest/__pycache__
+__pycache__
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.12)
-project(dfmodules VERSION 2.10.0)
+project(dfmodules VERSION 2.10.1)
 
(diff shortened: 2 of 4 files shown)


eflumerf (approved):
LGTM, though I might take the `new_` out of the file names, since they're already in a `drafts` directory.

PR #288, opened 2023-05-04
Eflumerf/integ test wib eth

eflumerf:


plugins/DataWriter.cpp
```diff
@@ -288,7 +288,7 @@ DataWriter::receive_trigger_record(std::unique_ptr<daqdataformats::TriggerRecord
     } //  if m_data_storage_is_enabled
   }
   
-  bool send_trigger_complete_message = true;
+  bool send_trigger_complete_message = m_running.load();
   if (trigger_record_ptr->get_header_ref().get_max_sequence_number() > 0) {
     send_trigger_complete_message = false;
     daqdataformats::trigger_number_t trigno = trigger_record_ptr->get_header_ref().get_trigger_number();
```
bieryAtFnal: The change from "true" to "m_running.load()" on line 291 is a nice catch, and I expect that it effectively prevents the sending of TriggerComplete messages after the DataWriter has received the Stop command when there is only one TriggerRecord sequence.  
However, when there is more than one TR sequence, it seems to me that "send_trigger_complete_message" can be reset to "true" by code within the "if" block that starts on line 292, even when the run has been stopped (as far as the DataWriter knows).  
Could/should the "if" statement on line 292 include a test that send_trigger_complete_message is "true" before it checks whether the current TR message completes a sequence?

eflumerf: I'm resolving this conversation, with commit f4904e3. Now when there are multiple sequences in the trigger, `send_trigger_complete_message` is set to false in the `else` case of the if statement checking for the last sequence. This means that if it was set to `false` by L291, it will remain false even if the current trigger record represents the last in the sequence.

bieryAtFnal (approved):
These changes look good to me, and in truth, I've added a few changes myself (which I've communicated to Eric).  
I've successfully run all of the tests on iceberg01.fnal.gov except the insufficient_disk_space_test, which I will get to, but maybe we go ahead with merging these changes in the meantime.
One comment for possible future work is that it might be useful to go through and remove the old WIB1/WIB2 (commented out) code at some point.

eflumerf:
I have also run all of the integration tests on `daq.fnal.gov`

PR #290, opened 2023-05-16
Add check on CPU count and free memory to long_window_readout_test

eflumerf:


bieryAtFnal:
I increased the minimum CPU count and free memory a bit, based on what I saw on mu2edaq13 (screenshot attached).
Screen Shot 2023-05-17 at 10 42 25 AM

bieryAtFnal (approved):
Looks great to me, thanks for adding this.
In addition to slightly increasing a couple of minimum sizes, I added the setting of the dfapp output_paths config parameter so that we can use locations other than "/tmp" for the output files.  (note that I needed to use the _integrationtest_ branch for multiple output files when testing a non-default data output location).

PR #293, opened 2023-05-31
ignore incoming TPSets of type kHeartbeat

glehmannmiotto:
This change in TPWriter avoids a crash of the application if the incoming TPSet is of type kHeartbeat.


roland-sipos (approved):
LGTM

PR #296, opened 2023-06-15
Removed the ignored_logfile_problems defined in the insufficient_disk…

bieryAtFnal:
…_space_test because they are now fully handled by the required_logfile_problems (based on recent changes in the integrationtest package)


eflumerf (approved):
LGTM

PR #298, opened 2023-06-26
Fix and re-enable TPG integration tests

eflumerf:
Now that WIBEth TPG is supported, re-enable the TPG portions of the integration tests, with appropriate updates.


eflumerf:
Note that this does *not* include planned changes for using a generated TPG file, those have been captured in https://github.com/DUNE-DAQ/dfmodules/issues/299

PR #305, opened 2023-07-19
Moved integtests to/from various repos to better reflect their intended scope

bieryAtFnal:
The `dfmodules` package mainly lost integtests as we moved several to the `daq-systemtest` repo...

This PR is related to similar ones in the `daq-systemtest` and `hsilibs` repos, but they are not strongly coupled, and they can be tested and reviewed independently.

A summary of the desired locations for various integtests is available in [this spreadsheet](https://docs.google.com/spreadsheets/d/1kMypxqOpZMB9_zY7XdJt7Fjuv2CheEVwsv7zHkWdPnE).


integtest/insufficient_disk_space_test.py
```diff
@@ -103,10 +103,7 @@
 conf_dict = config_file_gen.get_default_config_dict()
 conf_dict["daq_common"]["data_rate_slowdown_factor"] = data_rate_slowdown_factor
 conf_dict["readout"]["use_fake_data_producers"] = True
-#conf_dict["readout"]["default_data_file"] = "asset://?label=DuneWIB&subsystem=readout" # DuneWIB
-conf_dict["readout"]["default_data_file"] = "asset://?checksum=e96fd6efd3f98a9a3bfaba32975b476e" # WIBEth
```
eflumerf, line 107: I'm a little confused about why this line was removed...Is there a new default value for readout.default_data_file that makes the specification here redundant?

bieryAtFnal: From what I recall, when a fake data producer is used, there is no need for a playback-data file.

bieryAtFnal:
Closing this PR in favor of #306 which has all of these changes *and* includes a better organization of some `integrationtest` tools suggested by Eric.

PR #307, opened 2023-09-05
Integtest changes for new wib pattern generation

bieryAtFnal:
These changes are needed to keep the TPG parts of the integtests working after the latest changes to the WIB pattern generation.


eflumerf (approved):
LGTM

PR #309, opened 2023-09-07
Added script to run multiple integtests

bieryAtFnal:
This may be helpful as hopefully more people start running integtests.

The code isn't very elegant, but hopefully it is a reasonable start.


eflumerf:
Here is what I've been using, maybe there are some ideas that can be used...

[run_integration_tests.sh.txt](https://github.com/DUNE-DAQ/dfmodules/files/12551333/run_integration_tests.sh.txt)

bieryAtFnal:
Thanks Eric, for sending a copy of the script that you have been using.  I've incorporated some ideas and made some independent changes.  There are still improvements to be made, but I'm going to go ahead and merge this in order to get something into the release.

PR #312, opened 2023-09-15
Update CMakeLists.txt

dingp:
Remove `detchannelmaps` dependency in `daq_codegen` call.

This is to avoid warning messages like: 

`path does not exist: /cvmfs/dunedaq-development.opensciencegrid.org/nightly/NAB23-09-14/spack-0.20.0-gcc-12.1.0/spack-0.20.0/opt/spack/linux-almalinux9-x86_64/gcc-12.1.0/detchannelmaps-NAB23-09-14-2picxyo3hpxfy4mtmzujgeayn75nlq7r/lib64/detchannelmaps/cmake/../../../share/schema`


bieryAtFnal (approved):
This change gets rid of the warning messages, as expected, thanks!

PR #314, opened 2023-09-27
Increased the configured number of DFO/TRB trigger-record slots…

bieryAtFnal:
…in the multi_output_file_test.py integtest to avoid startup issues when occasional slowness in creating DataRequest Senders happens.

I first noticed TriggerInhibits when running the 3ru_3df integtest in the daq-systemtest repo, and I'm updating the value here as a precaution.


eflumerf (approved):
LGTM, integration tests pass

PR #316, opened 2023-10-17
Converted integtests to use arrays for TA and TC plugins

bieryAtFnal:
These change need to be tested and merged with correlated changes in daqconf ([PR 394](https://github.com/DUNE-DAQ/daqconf/pull/394)).


eflumerf (approved):
Integration tests succeeded

PR #325, opened 2023-12-19
Changed the bundle script to use the pytest return code for stop-on-failure

bieryAtFnal:
as suggested by Eric.


eflumerf (approved):
LGTM...now I need to figure out why I had a failure with the first test...

PR #326, opened 2024-01-11
Added a stop-on-skip option to the script that runs multiple integtests

bieryAtFnal:
This would be useful in situations in which all integtests are expected to pass and a skipped one would be unexpected.


eflumerf (approved):
Works, though it does raise the issue that most of the dfmodules regression tests do not work right now.

Also `insufficient_disk_space_test` is almost always skipped, unless the user went to some trouble to set up the correct ramdisk area.

eflumerf:
Opened #404 to deal with issues with both the integration tests and the integration test script. (Probably just needs to sync with daqsystemtest again)

wesketchum:
As it's approved (and it also looks ok to me), I'm going to go ahead and merge this in.

PR #328, opened 2024-02-07
Use source_id instead of src_id for all users of source IDs in config

eflumerf:


eflumerf:
These PRs should be treated as an ensemble:
https://github.com/DUNE-DAQ/coredal/pull/12
https://github.com/DUNE-DAQ/appdal/pull/22
https://github.com/DUNE-DAQ/dfmodules/pull/328
https://github.com/DUNE-DAQ/readoutlibs/pull/148
https://github.com/DUNE-DAQ/readoutmodules/pull/66

PR #330, opened 2024-02-29
Feature/oks

eflumerf:


plugins/SchemaUtils.hpp
gcrone, line 1: For the future we may want to consider making these classes inherit from the Jsonable schema class rather than have the converisons done here.

gcrone (approved):
Looks good to me!

PR #332, opened 2024-03-04
Update FragmentAggregator.cpp

glehmannmiotto:
moved debug TLOG

### DIFF
--- a/plugins/FragmentAggregator.cpp
+++ b/plugins/FragmentAggregator.cpp
@@ -109,7 +109,7 @@ FragmentAggregator::process_data_request(dfmessages::DataRequest& data_request)
                                                           data_request.trigger_number,
                                                           data_request.sequence_number));
     } else {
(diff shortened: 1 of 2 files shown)


plugins/FragmentAggregator.cpp
```diff
@@ -109,7 +109,7 @@ FragmentAggregator::process_data_request(dfmessages::DataRequest& data_request)
                                                           data_request.trigger_number,
                                                           data_request.sequence_number));
     } else {
-      TLOG() << "Send data request to " << uid_elem->second;
+      //TLOG() << "Send data request to " << uid_elem->second;
```
eflumerf: What about using `TLOG_DEBUG(30)` to make the message only print when we specifically enable it?

glehmannmiotto: fixed

eflumerf (approved):
LGTM

PR #333, opened 2024-03-12
Uses changes on branch with same name in appdal to have TRB fetch

eflumerf:
DataRequest connections generically from SmartDaqApplications.


glehmannmiotto (commented):
I don't think that it is true in general that any smart application which has a source ID is a potential fragment provider. This should be complemented by the check that the application exposes a DataRequest service.

eflumerf:
That's why I added a check against the presence of a DataRequest NetworkDescriptor at line 149

glehmannmiotto (approved):
LGTM

PR #336, opened 2024-03-25
Kbiery/tpswriter improvements

bieryAtFnal:
This PR is a response to [ehn1-operations-issues 30](https://github.com/DUNE-DAQ/ehn1-operations-issues/issues/30), "tpwriter errors on TimeSliceHeader dataset creation".  It has correlated changes in [daqconf](https://github.com/DUNE-DAQ/daqconf/pull/440) and [daqsystemtest](https://github.com/DUNE-DAQ/daqsystemtest/pull/85).

To help manage the error messages that are seen at EHN1, this PR adds support for a TPStreamWriter configuration parameter that can be used to lengthen the amount of time that the TPSW waits for a TimeSlice to be accumulated.  We believe that increasing this time will allow a wider variation of TPSet times to be accumulated in each TimeSlice in the TPStreamWriter, and this will avoid the error condition in which a TPSet arrives with a timestamp that is in a time interval (TimeSlice) that has already been written to the HDF5 file.

The new `daqconf` configuration input parameter is verbosely named `tp_accumulation_inactivity_time_before_write_sec`, and it has a default value of 1.0 seconds.  

The changes in this PR also

1. downgrade the annoying Error message to a Warning message
2. fix a bug in which a wide variation of TPSet timestamps at the beginning of a run can lead to nonsense slice numbers in the TimeSlices that are written to the HDF5 file
3. attempt to gracefully handle a moderate variation in TPSet timestamps at the beginning of a run.  If the variation cannot be gracefully handled, a warning message is printed out
5. fix a bug in which pending TimeSlices were not written out at the end of a run.

### DIFF
--- a/plugins/TPStreamWriter.cpp
+++ b/plugins/TPStreamWriter.cpp
@@ -158,36 +158,42 @@ TPStreamWriter::do_work(std::atomic<bool>& running_flag)
 
   TPBundleHandler tp_bundle_handler(m_accumulation_interval_ticks, m_run_number, m_accumulation_inactivity_time_before_write);
 
-  while (running_flag.load()) {
+  bool possible_pending_data = true;
+  while (running_flag.load() || possible_pending_data) {
     trigger::TPSet tpset;
     try {
       tpset = m_tpset_source->receive(m_queue_timeout);
       ++n_tpset_received;
       ++m_tpset_received;
+
+      if (tpset.type == trigger::TPSet::Type::kHeartbeat)
+        continue;
+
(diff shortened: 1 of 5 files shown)


bieryAtFnal:
Wes has provided feedback on these changes, and we've discussed possible additional improvements.  These include suppressing the ERROR message from the HighFive library and including an indication of how much out-of-time a tardy TPSet is.  We will look into those soon.

However, we believe that there are benefits from going ahead and merging the changes that have been made so far, for example doing a better job of including pending TimeSlices in the output files at end-run.  So, I will do the merges soon.

I believe that Wes has successfully tested these changes at EHN1.

PR #357, opened 2024-06-25
Johnfreeman/daq release issue379 renames

jcfreeman2:


eflumerf (approved):
Builds, no readoutlibs references remain in ldd of .so-files

PR #359, opened 2024-07-03
Fixed the file_index Attribute inside long-readout-window HDF5 files when one-event-per-file is enabled

bieryAtFnal:
Modified the HDF5DataStore so that it stores the correct file_index as an Attribute inside the HDF5 file. This required a bit of reworking of the file_index logic.

This is just a draft PR until I confirm expected behavior of various use cases...

### DIFF
--- a/plugins/HDF5DataStore.hpp
+++ b/plugins/HDF5DataStore.hpp
@@ -135,6 +135,7 @@ public:
 
     m_file_index = 0;
     m_recorded_size = 0;
+    m_current_record_number = std::numeric_limits<size_t>::max();
 
     if (m_operation_mode != "one-event-per-file"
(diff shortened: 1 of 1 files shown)


bieryAtFnal:
Marked this PR as ready-for-review (no longer draft).

I was over-thinking different use cases and got confused about the behavior that we want.  After some additional thinking, I've updated the changes to keep the general behavior that we've always had (with regards to when we close one file and open another one based on the size of the first file).

bieryAtFnal:
I've verified that these changes work fine with the updated daqsystemtest and dfmodules integtests for fddaq-v4.4.4.  (currently, those updated integtests are on branches...)

wesketchum (approved):

PR #372, opened 2024-08-12
State argument removed from execute_command

eflumerf:


glehmannmiotto (approved):
LGTM; there was a fix provided my @mroda88 on the TRB; to be checked that the merge works correctly.

PR #374, opened 2024-08-14
add application name and module name to the file name

### LINKED ISSUE #373: Names of HDF5 files in v5 minidaq system noticeably different than in v4
I'm not sure if this is the correct repository for filing this Issue, but I'll start here and be willing to move it, if needed.

Basically, my concern is that the filenames for raw data and TP stream HDF5 files from a simple minidaq test show a very different pattern than what we used previously, *and* the new pattern seems to have some undesirable features.

Here are three sample files from a v5 minidaq test, two produced by TRB/DataWriters and one produced by the TPStreamWriter:
```
test_raw_run000101_0000_default_20240814T164113.hdf5
test_raw_run000101_0000_default_20240814T164114.hdf5
test_raw_run000101_0000_default_tp_store_conf_20240814T164114.hdf5
```

Here are three sample files from a corresponding v4 minidaq test.
```
swtest_run000921_0000_dataflow0_datawriter_0_20240814T172121.hdf5
swtest_run000921_0000_dataflow1_datawriter_0_20240814T172121.hdf5
tpstream_run000921_0000_tpwriter_tpswriter_20240814T172122.hdf5
```

For reference, here are some sample filenames from NP04 running:
```
/data4/np04hd_raw_run028694_0009_dataflow6_datawriter_0_20240814T172224.hdf5
/data0/np04hd_tp_run028694_0018_tpwriter_tpswriter_20240814T171951.hdf5
```

My concerns about the v5 filenames include the following:

1.  The two TRB/DataWriter files are only distinguished by the timestamp.  Recall that this timestamp is only included as a safety check to avoid duplicate filenames, so we shouldn't be counting on it as the first level of defense for unique filenames.
2. The TPStream filename has a prefix of "test_raw" instead of "test_tp".
3. The TPStream filename has odd text in it.  What does "tp_store_conf" mean, and why is it in the filename?
4. There doesn't seem to be an indication of the process name in any of the filenames.  Maybe this is not important, but it is part of how filenames are distinguished in v4.x running.

glehmannmiotto:
This pull request should fix #373

### DIFF
--- a/include/dfmodules/DataStore.hpp
+++ b/include/dfmodules/DataStore.hpp
@@ -46,9 +46,10 @@
 #define DEFINE_DUNE_DATA_STORE(klass)                                                                                  \
   EXTERN_C_FUNC_DECLARE_START                                                                                          \
   std::unique_ptr<dunedaq::dfmodules::DataStore> make(const std::string& name,                                         \
-                                                      std::shared_ptr<dunedaq::appfwk::ModuleConfiguration> mcfg)      \
+                                                      std::shared_ptr<dunedaq::appfwk::ModuleConfiguration> mcfg,      \
+		  				      const std::string& mod_name)                                     \
   {                                                                                                                    \
-    return std::unique_ptr<dunedaq::dfmodules::DataStore>(new klass(name, mcfg));                                      \
+    return std::unique_ptr<dunedaq::dfmodules::DataStore>(new klass(name, mcfg, mod_name));                            \
   }                                                                                                                    \
   }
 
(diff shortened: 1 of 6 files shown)


bieryAtFnal:
Thanks for the fix.  With these changes, I now see

```
test_raw_run000111_0000_df-01_df-01-dw-0_20240817T185738.hdf5
test_raw_run000111_0000_df-02_df-02-dw-0_20240817T185737.hdf5
test_raw_run000111_0000_tp-stream-writer_tpwriter-4_20240817T185732.hdf5
```

If we are OK with "df-0x" appearing twice then things look good.  Are we OK with that?

glehmannmiotto:
The duplication of the application name comes from the code that sets the UID for the generated DAQ modules.
I would argue that there is no need to set the application name in the modules name, since the module lives only in the application scope. If we leave the naming as is, we will see this duplication also in ERS and opmon. Nevertheless, before changing this we need to check whether removing the application name would cause issues to the dataflow messaging system (and connectivity service)

glehmannmiotto:
can we decide on whether this fixed should be merged into develop or not?

PR #377, opened 2024-08-29
Opmon updates

mroda88:


eflumerf (approved):
LGTM, new metrics appear in output

PR #378, opened 2024-08-30
DataStoreConf now references a StorageDevice resource

eflumerf:
Test new functionality in HDF5Write_test.


eflumerf:
Ensemble with
https://github.com/DUNE-DAQ/dfmodules/pull/378
https://github.com/DUNE-DAQ/confmodel/pull/40
https://github.com/DUNE-DAQ/appmodel/pull/115

wesketchum:
I'm not sure about the quota thing ... do I have it right that this means a quota is set at configuration, and then a data writer can write up to that amount but after that it will give warnings that there isn't more space available?

- How does assigning that quota work? Does it actually look up the space on disk or just purely configuration? Is the configuration going to dynamically determine it's quota?
- For long-going runs, it could be a data writer writes lots of data, but because of the file management they are transferred and then deleted, and so there's no issue with space. We don't want to lock ourselves out of writing more data from a data writer if there's not actually a problem with writing more data, do we? But maybe I don't understand it.

eflumerf:
These are great points, and the quota was only implemented because it was implied by the configuration. However, in talking with Gordon, I realized that this was never actually a requested feature, so I'm backing it out.

PR #379, opened 2024-08-30
ta's are not stored in readout app

glehmannmiotto:


eflumerf (approved):
I generally lean towards deleting code and letting Git handle it, but this is fine for a quick fix.

glehmannmiotto:
code removed and better check introduced on tp generation.

eflumerf:
I don't think you pushed, since this appears to be unchanged since I commented last.

eflumerf (approved):
Unit tests and a simple test-session run look good

PR #387, opened 2024-10-22
Added a get_receiver() call for the DataRequest input in the FragmentAggregator init() method.

bieryAtFnal:
This is to get the receiver registered early, before we install the callback at Start time, so that senders can find it when they look for it.

This helps avoid the warning messages that we have been seeing in the minimal_system_quick_test in which the TRB complains about not being able to send DataRequests at the start of the run.

I'm including the addition of a debug statement in this PR that I found useful several weeks ago when debugging a different problem.

To test this change, I locally modified a copy of the minimal_system_quick_test to have no wait between start and enable-triggers.  That modified test fails fairly often on daq.fnal.gov without the code change in the PR, and has not failed yet in my testing with this code change.


eflumerf (approved):
Ran minimal_system_quick_test as described, saw that error is fixed.

PR #395, opened 2024-11-26
Added checks in do_start() methods to verify that data senders are ready to send data

bieryAtFnal:
...in DFModule, TRBModule, and FragmentAggregatorModule.

The background for these changes is that we noticed that the automated regression tests that include TPG (trigger primitive generation)  in the `daqsystemtest` package would sometimes have Warning messages in the DFO and/or MLT log files saying that triggers were inhibited or had been assigned to TRBuilders that were already fully busy.  These warning messages would only happen at the start of the first run in a DAQ session.

Debugging this by looking at TRACE messages from various Dataflow software modules, it became clear that there were slight delays in sending the first several messages between Dataflow and Readout DAQ modules at the start of the first run.  Another clue to the cause of this problem was the observation that the Connectivity Service was running somewhat slower than normal when the "inhibit" warning messages appeared.

The proposed solution is to cause the various IOManager Sender objects In the relevant DAQModules to pre-fetch the needed destination information from the ConnectivityService.  This is done by calling the "is_ready_for_sending()" method of the Sender objects in the DAQModule do_start() method.

To reproduce the problem and validate the changes in this PR, we can use the following steps:

- create a software area based on a recent nightly build, such as described [here](https://github.com/DUNE-DAQ/daqconf/wiki/Setting-up-a-fddaq%E2%80%90v5.3.0-development-area).
- check out and install a temporary version of the connectivityserver software in this software area.  This temporary version of the ConnectivityService includes an artificial delay in its get_connection operation to make it appear as if it is running slowly.  Here are suggested steps to do this:
    - `cd $DBT_AREA_ROOT`
    - `git clone https://github.com/DUNE-DAQ/connectivityserver.git -b develop`
    - `cd connectivityserver`
    - `git checkout kbiery/artificial_get_connection_delay`
    - `pip install -U .`
- run the TPG part of the 3ru_3df_multirun_test
    - `cd $DBT_AREA_ROOT/sourcecode/daqsystemtest/integtest`
    - `pytest -k TPG -s 3ru_3df_multirun_test.py`
- note the warning messages in logfiles that cause the integtest to fail
- update the dfmodules software to include the changes in this PR
    - `cd $DBT_AREA_ROOT/sourcecode/dfmodules`
    - `git checkout kbiery/check_senders_ready`
    - `cd $DBT_AREA_ROOT`
    - `dbt-build -j 12 ; dbt-workarea-env`
- re-run the TPG part of the 3ru_3df_multirun_test
    - `cd $DBT_AREA_ROOT/sourcecode/daqsystemtest/integtest`
    - `pytest -k TPG -s 3ru_3df_multirun_test.py`
- note that the integtest now runs cleanly/successfully
- restore the ConnectivityService code to its default version
    - `cd $DBT_AREA_ROOT/connectivityserver`
    - `git checkout develop`
    - `pip install -U .`
    - `cd $DBT_AREA_ROOT`

### DIFF
--- a/plugins/DFOModule.cpp
+++ b/plugins/DFOModule.cpp
@@ -79,6 +79,9 @@ DFOModule::init(std::shared_ptr<appfwk::ModuleConfiguration> mcfg)
     if (con->get_data_type() == datatype_to_string<dfmessages::TriggerInhibit>()) {
       m_busy_sender = iom->get_sender<dfmessages::TriggerInhibit>(con->UID());
     }
+    if (con->get_data_type() == datatype_to_string<dfmessages::TriggerDecision>()) {
+      m_trb_conn_ids.push_back(con->UID());
+    }
   }
 
   if (m_token_connection == "") {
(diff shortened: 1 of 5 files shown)


eflumerf (approved):
Suggested test regimen shows issue and resolution.

PR #416, opened 2025-06-05
Add verbose error

mroda88:
When the fragment aggregator fails to send, the issue is reported as a warning which is not descriptive because we don't have any information about the what is lost. We only know what error was reported by the iomanager. This is the typical example of what is seen. 

Screenshot From 2025-06-04 12-54-30

This PRs turns the warning into an error, as there is data loss and it also adds a dedicated ERS message to specify what was lost. 
This is a screenshot of the message that we see after the PR. 

image


bieryAtFnal (approved):
LGTM.  I locally modified the FragAgg code to occasionally produce the updated error message in a fake-data-replay system, and it looked good.  I also successfully ran the regression tests in the daqsystemtest repo.

mroda88:
Great, I'll merge and add to the local build in EHN1

PR #419, opened 2025-06-12
FragmentAggregator improved ers

MRiganSUSX:
Improving warning message when sender->send fails (request by @mroda88).

- passes unit test
- passes integ test bundle
- tested locally by manually triggering the catch block:
image

### DIFF
--- a/src/dfmodules/CommonIssues.hpp
+++ b/src/dfmodules/CommonIssues.hpp
@@ -70,9 +70,9 @@ ERS_DECLARE_ISSUE_BASE(dfmodules,
  */
 ERS_DECLARE_ISSUE(dfmodules,            ///< Namespace
                   DRSenderLookupFailed, ///< Issue class name
-                  "Unable to determine the Data Request message sender for SourceID [" << sid
-                  << "]. No DataRequest will be sent to this data source for run/trigger/sequence number "
-                  << runno << "/" << trigno << "/" << seqno << ".",
+                  "Unable to determine the Data Request message sender for SourceID ["
+                    << sid << "]. No DataRequest will be sent to this data source for run/trigger/sequence number "
+                    << runno << "/" << trigno << "/" << seqno << ".",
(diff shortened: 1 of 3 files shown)


MRiganSUSX:
@mroda88 
Few questions:
- do we want this to remain a `warning`
- are we sure the `sender->send` throws? (otherwise what is the point of this catch)
- the `data_request.data_destination` seems to be an empty string

src/dfmodules/CommonIssues.hpp
```diff
@@ -87,6 +87,22 @@ ERS_DECLARE_ISSUE(dfmodules,         ///< Namespace
...
+ * @brief Data Request send failed
+ */
+ERS_DECLARE_ISSUE(dfmodules,          ///< Namespace
+                  DRSenderSendFailed, ///< Issue class name
+                  "Failed to send Data Request message. Run: "
+                    << runno << ", trigger: " << trigno << ", trigger timestamp: " << triggerts << ", sequence: "
+                    << seqno << ", destination: " << datadest << ", requested component: " << sid << ".",
+                  ((daqdataformats::run_number_t)runno)      ///< Message parameters
+                  ((daqdataformats::trigger_number_t)trigno) ///< Message parameters
+                  ((daqdataformats::timestamp_t)triggerts)   ///< Message parameters
```
mroda88: I think we can slim down and remove the trigger time. It's not necessary to uniquely identify the data request.

mroda88:
For your questions. 
 - no, it must be turned into an error because it's data loss (eventually)
 -  We are sure, it can throw
 - I think we can simplify and get rid of the destination, we already have the SourceId, that's enough.

MRiganSUSX:
- changed to error
- removed parts you mentioned

test:
image

otherwise passed unit tests and integtests bundle.

mroda88:
great. Now if you could also bump the patch version, we can merge that in too. And we will just need to tag.

mroda88 (approved):
Great, thanks

PR #420, opened 2025-06-16
Adding discarded TPs metric

xinyue-uoft:
This pull request introduces new metric to track TPs that are discarded by the `TPStreamWriterModule`. This change adds two new fields to `opmon::TPStreamWriter`.

### Discarded Tardy TPs

The current counter introduced keeps track of the incremental and total number of TPs discarded when exception arise and 'IgnorableDataStoreProblem' is true. In other words, only those associated with TardyTPsDiscarded warning is counted. In terms of code organization, the additional metric is analogous to the already present tps received and written.

### Tests
- passes 'dbt-unittest-summary'
- passes 'integtest/minimal_system_quick_test'

### DIFF
--- a/plugins/TPStreamWriterModule.cpp
+++ b/plugins/TPStreamWriterModule.cpp
@@ -75,8 +75,10 @@ TPStreamWriterModule::generate_opmon_data() {
   info.set_tpsets_with_tps_received(m_tpsets_with_tps.exchange(0));
   info.set_tps_received(m_tps_received.exchange(0));
   info.set_tps_written(m_tps_written.exchange(0));
+  info.set_tps_discarded(m_tps_discarded.exchange(0));
   info.set_total_tps_received(m_total_tps_received.load());
   info.set_total_tps_written(m_total_tps_written.load());
+  info.set_total_tps_discarded(m_total_tps_discarded.load());
   info.set_tardy_timeslice_max_seconds(m_tardy_timeslice_max_seconds.exchange(0.0));
   info.set_timeslices_written(m_timeslices_written.exchange(0));
   info.set_bytes_output(m_bytes_output.exchange(0));
@@ -276,6 +278,9 @@ TPStreamWriterModule::do_work(std::atomic<bool>& running_flag)
               else {sid_list << ",";}
               sid_list << frag_ptr->get_element_id().to_string();
             }
+	    size_t number_of_tps_discarded = (timeslice_ptr->get_sum_of_fragment_payload_sizes() / sizeof(trgdataformats::TriggerPrimitive));
+	    m_tps_discarded += number_of_tps_discarded;
+	    m_total_tps_discarded += number_of_tps_discarded;
             ers::warning(TardyTPsDiscarded(ERS_HERE,
                                            get_name(),
                                            sid_list.str(),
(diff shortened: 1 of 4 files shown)


mroda88 (changes requested):
The total counter should be reset to 0 in `do_start`.

xinyue-uoft:
Tests

1. passes 'dbt-unittest-summary'
2. passes 'integtest/minimal_system_quick_test'

xinyue-uoft:
Renamed `number_of_tps_written` to `number_of_tps`, this is because it can be used again in handling discarded tps, therefore no need to do the same calculation at two different placed. Its declaration is moved out of the try block inside the do while loop. This is due to the try block and catch block not in the same scope, so that local variable declared and set in try can not be accessed in catch.

Fixed the behaviour that number of discarded TPs is not recorded when `m_warn_user_when_tardy_tps_are_discarded` is `false`. Whether we throw a warning does not change the fact that TP was discarded.

Tests

1. passes 'dbt-unittest-summary'
2. passes 'integtest/minimal_system_quick_test'

xinyue-uoft:
Tests

1. passes 'dbt-unittest-summary'
2. passes 'integtest/minimal_system_quick_test'

plugins/TPStreamWriterModule.cpp
```diff
@@ -245,29 +248,29 @@ TPStreamWriterModule::do_work(std::atomic<bool>& running_flag)
       size_t retry_wait_usec = 1000;
       do {
         should_retry = false;
+	size_t number_of_tps = 0;
         try {
+	  number_of_tps = (timeslice_ptr->get_sum_of_fragment_payload_sizes() / sizeof(trgdataformats::TriggerPrimitive));
```
mroda88: this outside the try block.

plugins/TPStreamWriterModule.cpp
```diff
@@ -245,29 +248,29 @@ TPStreamWriterModule::do_work(std::atomic<bool>& running_flag)
...
-          usleep(retry_wait_usec);
-          retry_wait_usec *= 2;
+	  usleep(retry_wait_usec);
+	  retry_wait_usec = std::min(retry_wait_usec * 2, 1000000UL);
         } catch (const IgnorableDataStoreProblem& excpt) {
           int timeslice_number_diff = largest_timeslice_number - timeslice_ptr->get_header().timeslice_number;
           double seconds_too_late = m_accumulation_interval_seconds * timeslice_number_diff;
           m_tardy_timeslice_max_seconds = std::max(m_tardy_timeslice_max_seconds.load(), seconds_too_late);
+	  m_tps_discarded += number_of_tps;
+	  m_total_tps_discarded += number_of_tps;
```
mroda88, line 272: We need the increments also in the DataWriting problem catch, a few lines down the code.

mroda88 (dismissed):

xinyue-uoft:
Current behaviour is to increment tps discarded and total tps discarded counters when all of these are true: 1. Exception occurs. 2. The exception is not a `RetryableDataStoreProblem`.

Tests:

1. passes 'dbt-unittest-summary'
2. passes 'integtest/minimal_system_quick_test'

MRiganSUSX (approved):
LGTM.

The code is correct.
Passes unit-tests.
Passes minimal_system_quick_test.

@mroda88 has also implemented a plot for these metrics as part of the `Trigger Primitives` dashboard, and we see the metrics to be published/updated.

xinyue-uoft:
Updated dfmodules VERSION 4.1.2 to dfmodules VERSION 4.1.3.

PR #423, opened 2025-06-24
Make the TRBModule callback-based

eflumerf:


mroda88 (approved):
This was tested at EHN1, adding stress tests too. Dashboard of the test is http://np04-srv-017:31023/d/v5_4_0-overview/daq-overview?orgId=1&var-influxdb=f68a27a2-f292-478c-8093-2a4adad3be1b&var-postgresql=c7e9e333-25c9-49e7-b209-b9492f70d419&var-session=mrigan-np02-mrigan-stress-test&var-controller=root-controller&from=1753783207554&to=1753791042586 The first run was with cubic congestion algorithm (which is not the desired one) and the second run was with congestion algorithm set to BBR. 

The performance of the new TRB is the same as the old one with the BBR congestion algorithm, and even better with the cubic algorithm. So this is desirable for v5.4.0 as discusse in the SWIT meeting today.

PR #425, opened 2025-07-17
Fixed the indentation in the bundle script (no substantive changes, …

bieryAtFnal:
…only formatting).

The indentation of the lines of code in the script has become inconsistent over time, and this change is just to make it consistent again.


eflumerf (approved):
Turned on "Hide Whitespace", saw that the diff ended up being two lines removed. Approved.

PR #431, opened 2025-08-21
Various fixes to the TRBModule

eflumerf:
Only run bookkeeping periodically (currently every 1s).
Only call check_stale_requests from Fragment callback. 
Use C++17 map::extract to reduce time-under-lock in extract_trigger_record. 
Fix message printed when releasing incomplete TriggerRecord to only report Stop status when it is actually stopping.
Use `steady_clock` instead of `high_resolution_clock` for TriggerRecord timeouts.

### DIFF
--- a/plugins/TRBModule.cpp
+++ b/plugins/TRBModule.cpp
@@ -331,7 +331,7 @@ TRBModule::fragments_callback(std::unique_ptr<daqdataformats::Fragment>& temp_fr
 {
 
   auto start_time = std::chrono::steady_clock::now();
-  
+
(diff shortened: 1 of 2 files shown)


mroda88 (approved):
This was tested on a long run over the weekend. We've been running with TPG for the past 24 h and all seems stable

eflumerf:
More information: https://indico.fnal.gov/event/70916/contributions/322458/attachments/190902/263929/TRBModule%20Update%20%28PR%20%23431%29.pptx

PR #439, opened 2025-09-02
Add TRMonRequestorModule

eflumerf:
This module periodically requests TriggerRecords from the TRB, for use in DQM setups.


alessandrothea (approved):
LGTM. Tested at NP02

Issue #447, opened 2025-09-09
TRExtractor fixes


mroda88:
While using TRExtractor at EHN1 I noticed two issues:

- [x] Failed start at second start. I tried a the following transition sequences:
       start -> stop -> start (without reconfiguring). In the second run the TRExtractor was not sending request

- [x] The TRExtractor monitoring does not report the number of the received TR, it would be nice to add it

eflumerf:
For your second bullet, the TRMonRequestor does not receive the TriggerRecords. You should look at the DataWriter "records_received" metric from the TRMonReqApplication.

mroda88:
DataWriter plots has been added in the DQM dashboard with https://github.com/DUNE-DAQ/grafana-dashboards/pull/112. With this the Issue can be closed.

PR #448, opened 2025-09-09
Fixes for the TRMonRequestorModule when running multiple runs in a session

eflumerf:


eflumerf:
Resolves #447

mroda88 (approved):
Tested at EHN1 with 13 October nightly build. It works.

PR #450, opened 2025-09-24
Updated the use of the TimeSync message in FakeDataProdModule to use …

bieryAtFnal:
…SourceID::id instead of the process ID.

## Description

These changes are part of a set of changes that are described in DUNE-DAQ/daq-deliverables#188.  Overall, the goal is to eliminate the use of Linux process IDs in identifying the source of TimeSync messages (to avoid problems with their use in certain environments) and use DAQ SourceID information instead.

Please see the `daq-deliverables` Issue for more details, including the full set of repositories that have correlated changes.

## Testing Suggestions

Please see DUNE-DAQ/daq-deliverables#188.


eflumerf (approved):
With other repos with this branch, ran integration test suite

PR #451, opened 2025-09-29
For tests that use HSI Fragments, set the size min/max using the TC type

### LINKED ISSUE #256: Reduce range in size of HSI fragments in integtests?
At the moment, the allowed sizes of HSI fragments in dfmodules/integtest tests are [72,100].  The lower bound corresponds to empty fragments, which we needed to support at various times when we would occasionally get empty HSI fragments.
The point of this Issue is to ask the question whether those problems have been fixed and we can now expect a narrow range in HSI fragment sizes, essentially [100,100].
If so, this Issue would also request that such a change be made in the integtests.

eflumerf:
(copied from daqsystemtest)

Resolves #256

### DIFF
--- a/integtest/hdf5_compression_test.py
+++ b/integtest/hdf5_compression_test.py
@@ -103,12 +103,16 @@ triggerprimitive_frag_params = {
     "min_size_bytes": 72,
     "max_size_bytes": 1032,
 }
+# 03-Jul-2025, KAB: changing the default max size from 72 to 100 to handle cases in which there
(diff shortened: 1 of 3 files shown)


bieryAtFnal:
`multiple_data_writers_test.py` doesn't need this additional sophistication since it only uses kRandom triggers.  Can we back out the `frag_sizes_by_TC_type` changes in that case?
(otherwise, things are looking good)

bieryAtFnal (approved):
LGTM.  Thanks!

PR #454, opened 2025-10-17
Updated the WIBEth binary file used in the disabled_output_test.

bieryAtFnal:
## Description

DUNE-DAQ/daqsystemtest#245 requests that we update our regression tests, etc. to use recent data-replay files from the _assets_ system.

In this repo, that translated to changing the checksum of the asset file that is used in the `disabled_output_test`.  The new file is `wib_link_67.bin`, which was recently created by Eric.

*** Please note that we changed the intention of this PR.  Instead of updating the default frames file, we removed the specification of the frames file in favor of a comment that talks about default values.

## Testing Suggestions

There are suggestions in DUNE-DAQ/daqsystemtest#245 for testing all of the asset-file-update changes together.  In addition, the change in this repo could be tested independently by including the `kbiery/asset_file_refresh` branch of this repo in a software area created from a recently nightly build, and running the `disabled_output_test`.

### DIFF
--- a/integtest/disabled_output_test.py
+++ b/integtest/disabled_output_test.py
@@ -90,7 +90,7 @@ conf_dict.dro_map_config.n_streams = number_of_data_producers
 conf_dict.op_env = "integtest"
 conf_dict.session = "disabled"
 conf_dict.tpg_enabled = False
-conf_dict.frame_file = "asset://?checksum=e96fd6efd3f98a9a3bfaba32975b476e"  # WIBEth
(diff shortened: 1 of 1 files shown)


eflumerf (approved):
Question: If this is the default frame file, should it be specified at all?

bieryAtFnal:
I've removed the specification of the frame_file in favor of a comment that describes a little about what happens in the background.

bieryAtFnal:
As suggested by Eric, I removed the setting of the default frames file.

PR #469, opened 2025-12-29
FakeDataProd: Use TimestampEstimatorSystem instead of nanosecond timestamps 

### LINKED ISSUE #391: FakeDataProd uses nanosecond timestamps
It should use the DUNE 65 MHz clock tick timestamp

eflumerf:
when creating TimeSync messages.

# Description

Resolves Issue #391 by having FakeDataProdModule use TimestampEstimatorSystem to produce timestamps using the DUNE-DAQ clock (getting the clock frequency from the Session's DetectorConfig object). These timestamps are used to create TimeSync messages only; the module uses trigger timestamps for any produced data Fragments.

## Type of change

## Testing checklist

- [x] Unit tests pass (e.g. `dbt-build --unittest`)
- [x] Minimal system quicktest passes (`pytest -s minimal_system_quick_test.py`)
- [x] Full set of integration tests pass (`daqsystemtest_integtest_bundle.sh`)

I believe the only possible change is in `fake_data_producers_test.py`, which passed in my testing.

### DIFF
--- a/plugins/FakeDataProdModule.cpp
+++ b/plugins/FakeDataProdModule.cpp
@@ -11,6 +11,7 @@
 
 #include "appmodel/FakeDataProdModule.hpp"
 #include "confmodel/Connection.hpp"
+#include "confmodel/DetectorConfig.hpp"
 #include "dfmessages/Fragment_serialization.hpp"
 #include "dfmessages/TimeSync.hpp"
 #include "iomanager/IOManager.hpp"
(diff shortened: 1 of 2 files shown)


bieryAtFnal (approved):
These changes look good to me.  However, the `fake_data_producer_test` in `daqsystemtest` needed to be changed as a result.  That change is in DUNE-DAQ/daqsystemtest#288.

PR #474, opened 2026-01-22
Update usage of FragmentErrorBits to FragmentStatusBits

eflumerf:
# Description

Main PR: https://github.com/DUNE-DAQ/daqdataformats/pull/78
Issue: https://github.com/DUNE-DAQ/daqdataformats/issues/66

### DIFF
--- a/plugins/FakeDataProdModule.cpp
+++ b/plugins/FakeDataProdModule.cpp
@@ -193,7 +193,7 @@ FakeDataProdModule::process_data_request(dfmessages::DataRequest& data_request)
   data_fragment_ptr->set_trigger_number(data_request.trigger_number);
   data_fragment_ptr->set_run_number(m_run_number);
   data_fragment_ptr->set_element_id(m_sourceid);
-  data_fragment_ptr->set_error_bits(0);
+  data_fragment_ptr->set_status_bits(0); // Do we want to have a kFakeData status bit?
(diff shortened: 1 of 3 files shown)


plugins/FakeDataProdModule.cpp
```diff
@@ -193,7 +193,7 @@ FakeDataProdModule::process_data_request(dfmessages::DataRequest& data_request)
   data_fragment_ptr->set_trigger_number(data_request.trigger_number);
   data_fragment_ptr->set_run_number(m_run_number);
   data_fragment_ptr->set_element_id(m_sourceid);
-  data_fragment_ptr->set_error_bits(0);
+  data_fragment_ptr->set_status_bits(0); // Do we want to have a kFakeData status bit?
```
wesketchum: Can me either just answer this question, or make an issue to ask/discuss/answer this question rather than just leave it as a comment? I worry if we leave it here, it will stay here for a decade ...

wesketchum (changes requested):
Code looks good. See comment on the comment.

plugins/FakeDataProdModule.cpp
eflumerf: https://github.com/DUNE-DAQ/dfmodules/issues/479

wesketchum (approved):

PR #491, opened 2026-06-24
Modified integtests to make use of new utility functions for setting trigger config params

bieryAtFnal:
# Description

These changes make use of the changes made in the `integrationtest` repo to provide utility functions that set various trigger parameters for integration tests (DUNE-DAQ/integrationtest#161).

There are instructions for testing these changes in DUNE-DAQ/integrationtest#161, and these changes need to be tested and merged together the with changes in the `integrationtest` repo.

## Type of change

- [x] Optimization (non-breaking change that improves code/performance)

## Testing checklist


integtest/large_trigger_record_test.py
```diff
@@ -133,19 +126,10 @@
...
-            "trigger_rate_hz": trigger_rate,
-            "candidate_backshift_ts": 0,
-            "candidate_window_before_ts": 2.5 * readout_window_time_before,
-            "candidate_window_after_ts": readout_window_time_after
-        },
-    )
-)
+utility_functions.set_RTCM_trigger_params(oversize_conf, trigger_rate=trigger_rate,
+                                          readout_window_backshift_ticks=0,
+                                          readout_window_before_ticks=2.5*readout_window_time_before,
```
MRiganSUSX: `readout_window_time_before` is hinted to be `int` (as it should) in https://github.com/DUNE-DAQ/integrationtest/blob/24fce0f387d7609946277b77ee39612213cc65f1/src/integrationtest/utility_functions.py#L102 but used here with float multiplication (2.5).
While this won't cause any problems at runtime, it can break some python type checkers (as int is compatible with float but not the other way around). 
I'd suggest casting to int here as:
```
int(2.5*readout_window_time_before),
```

MRiganSUSX (commented):
Just one comment here.

integtest/large_trigger_record_test.py
bieryAtFnal: Done, thanks.

MRiganSUSX (approved):
LGTM.

PR #497, opened 2026-07-10
Johnfreeman/file formats refactor

### LINKED ISSUE #496: [Feature]: Refactor of `DataStore` and its implementations
### Description

This can be considered part of the general File Formats review which Wes describes in https://gist.github.com/wesketchum/87312f5c722ae070b7c3f8109af78318. The main focus here is to take `DataStore` and its only implementation, `HDF5DataStore`, and separate file format (hdf5-) specific parts from non-file-format specific parts, creating reusable code which can be used by file-focused `DataStore` implementations. 

### Potential impact radius

Small/Isolated

### Reason for change

_No response_

### Suggested implementations

_No response_

### Testing suggestions

_No response_

### Anything else?

_No response_

jcfreeman2:
# Description

Addresses issue #496 

As discussed during my talk at the Dataflow meeting on Wednesday morning (Jul-8-2026), I'm filing a draft PR which (with a couple small exceptions which I'll describe below) is a refactor of `dfmodules`. The main focus of the refactor has been to separate the file-format-specific parts of `HDFDataStore` (e.g., constructing `hdf5libs`' `HDF5RawDataFile` class directly, or hardwiring in the `hdf5` extension to the output filename) from the file-format-agnostic parts (e.g., deciding when to close a file based on the number of bytes written to it). 

Note that there are some finishing touches I've left out - e.g., reordering `private` vs. `protected` vs. `public` declarations to adhere to our standards, sticking implementation details in `*.hxx` files, etc. The logic there is that by not doing that, it's easier to compare the intermediate base class I've created, `FileDataStoreImpl`, with the original `HDFDataStore` on the `develop` branch.

Some of the changes made include:
* Removing `HDF5FileUtils.hpp` and its unit tests. `#include`d in `HDF5DataStore` on the `develop` branch, it was last touched in 2021 and none of its functions are used. 
* Moving the code in `HDF5DataStore` which doesn't concern `hdf5` specifically into an intermediate base class, `FileDataStoreImpl`, which other file-format-specific implementations can inherit from. How to do this is described in the comment at the top of `FileDataStoreImpl`, but you can also see this "in action" by looking at `plugins/HDF5DataStore.hpp`
* There's also a "stub" implementation of `DataStore`, `StubDataStore` which inherits from `FileDataStoreImpl` and writes a very simple one line output to a text file when its `write` function is called for either trigger records or time slices
* Unit tests for `StubDataStore` are also added in `StubWrite_test`; they're basically the `HDF5DataStore` unit tests with `hdf5`-specific stuff stripped out, and are de-facto unit tests for `DataStoreImpl`. 

Some of the decisions I made in this Draft PR I imagine (hope) are uncontroversial - getting rid of a no-longer-used header, refactoring the code from `HDF5DataStore` when it was moved into `FileDataStoreImpl` to avoid Don't Repeat Yourself, making logically-constant member data declared `const`, etc. 

Others are open to discussion; I'll point some out here:
* On the `develop` branch, `hdf5libs::HDF5RawDataFile::write` for a timeslice can throw a `TimeSliceException`. However, because I don't want every file-format-specific class to have to supply its own `TimeSliceException` (which would then need to be supplied as a template parameter to `FileDataStoreImpl`) I require that the class which satisfies the `FileHandleConcept` have a boolean telling you if you passed it a duplicate time slice. 
* I removed the section of the `FileDataStoreImpl` constructor where if the free space safety factor is below 1.1 it gets forced to 1.1. I did so thinking the configuration should tell you straightforwardly what's going on, and if we want to set a lower limit, have that enforced in the configuration or at least get the constructor to throw rather than continuing on with a different value than the one it was given
* If the output path is invalid, a warning is issued, and only on the start transition (when `prepare_for_run` is called) is an exception thrown in this case. Curious why we don't just throw from the constructor?
* I left `HDF5Write_test` with its original name, though it could be called `HDFDataStore_test`. Perhaps this raises the question, though, as to whether `DataWrite` wouldn't be a more accurate name than `DataStore`?
* I know we discussed this at Dataflow but forget what answer was converged on - right now the run number is passed to `DataStore::finish_with_run` but `HDFDataStore`'s implementation on the develop branch (and hence `FileDataStoreImpl`'s implementation on this feature branch) don't use it. Do we want to keep this in the signature? _Should_ we be using it right now?
* In keeping with how it's done on the `develop` branch's `HDF5DataStore::write` function for the time slice - and the spirit of "this is just a refactor" - `FileDataStoreImpl::write` will check for duplicate time slices _after_ determining whether to open a new file. Should this be before? 

Also, moving forward, things to think about:
* Kurt raised questions about if we want to adjust the logic we use for when to open a new file
* Marco mentioned at Dataflow this week that the model of passing the configuration as a template parameter to the implementation of `DataStore` may not work long term, and that we'd need to literally just supply the instance of the `ConfigurationManager`; this is an issue that goes beyond the refactor and indeed beyond `DataStore`. 
* In `FileDataStoreImpl::open_file_if_needed`, `m_file_handle.reset();` is wrapped in a `try-catch` block. However by default destructors are `noexcept`, meaning no throw would ever make it into the `catch`.
 
To test, of course, check that this branch builds (it should be done alongside branches of the same name in `hdf5libs` and `trigger`, though in the case of `trigger` we're talking about a one-line change because I switched an argument in `hdf5libs` from a pointer to a reference). 

The latest test build for this branch is `DFRFD_DEV_260711_A9`, which used commit `84843de915bc51` (`Sat Jul 11 10:12:31 2026 -0500`). 

## Type of change

- [ ] Documentation (non-breaking change that adds or improves the documentation)
- [ ] New feature or enhancement (non-breaking change which adds functionality)
- [X ] Optimization (non-breaking change that improves code/performance)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (whatever its nature)

## Testing checklist

- [X ] Unit tests pass (e.g. `dbt-build --unittest`)
- [X ] Minimal system quicktest passes (`pytest -s minimal_system_quick_test.py`)
- [X ] Full set of integration tests pass (`dunedaq_integtest_bundle.sh`)
- [ ] Python tests pass if applicable (e.g. `python -m pytest`)
- [ ] Pre-commit hooks run successfully if applicable (e.g. `pre-commit run --all-files`)

_Comments here on the testing_

## Further checks

- [X ] Code is commented where needed, particularly in hard-to-understand areas
- [X ] Code style is correct (`dbt-build --lint`, and/or see https://dune-daq-sw.readthedocs.io/en/latest/packages/styleguide/)
- [X ] If applicable, new tests have been added or an issue has been opened to tackle that in the future.
  (Indicate issue here: # (issue))


eflumerf:
Instead of Stub*, maybe something along the lines of SummaryText* ?

include/dfmodules/FileDataStoreImpl.hpp
```diff
@@ -0,0 +1,613 @@
...
+ * An near-complete file output focused implementation of the DataStore interface which
+ * nonetheless templatizes the representation of the data file (e.g.,
+ * hdf5, Root, text, etc.) and leaves a few implementation
+ * details (e.g., the physical closing and reopening of data files)
+ * to a derived class.
+ *
+ * To create a full implementation of DataStore, perform the following steps:
+ *
+ * 1) Create a class which satisfies the constraints expressed in the
+ * FileHandleConcept below - hdflibs' HDF5RawDataFile is an example
```
eflumerf: Refer to Stub here as the simpler example

include/dfmodules/FileDataStoreImpl.hpp
```diff
@@ -0,0 +1,613 @@
...
+  { const_file_handle.get_total_file_size() } -> std::convertible_to<size_t>;
+};
+
+/**
+ * @brief FileDataStoreImpl contains functionality you'd generally
+ * want in a data store which writes to a file irrespective of file
+ * type
+ */
+  template <FileHandleConcept FileHandleClass, // E.g., hdf5libs::HDF5RawDataFile
+	    typename DataStoreConf> // E.g., appmodel::DataStoreConf
```
eflumerf: I know we had talked about removing the Conf template parameter using inheritance...

jcfreeman2:
Suggestions from @eflumerf have been implemented:
-Removal of configuration as template parameter to `FileDataStoreImpl` in favor of usage of OKS superclass inheritance
-Renaming of proof-of-concept implementation of `DataStore` from `StubDataStore` to `SummaryTextDataStore`

Also note that `SummaryTextDataStore` provides an example of how someone can use OKS inheritance for file-format-specific variables (`get_derivval()`)

jcfreeman2:
This is to be merged in concert with the feature branches of the same name in `hdf5libs` and `appmodel`. 

Also: In this last commit I rearranged the order of declarations for `FileDataStoreImpl` and moved some implementation details into a `FileDataStoreImpl.hxx` file; note then that for side-by-side comparisons with `develop` it may be easier to look at the penultimate commit on this feature branch.

jcfreeman2:
Concerning the failing single repo CI workflow: the linter complaints remaining either (1) have to do with code elsewhere, (2) appear to be a false positive (complaint about missing copyright) or (3) I'm disregarding (`std::make_unique`'s advantages aren't really that impressive).

jcfreeman2:
Note in response to Eric's comment on the PR for the `johnfreeman/file_formats_refactor` branch of `appmodel` (https://github.com/DUNE-DAQ/appmodel/pull/294) I've now added a `datastore_derived.schema.xml` file to this PR. No change is now needed for `appmodel`, so I've deleted the `johnfreeman/file_formats_refactor` branch there. The PRs, then, for repos with `johnfreeman/file_formats_refactor` branches besides this PR are:
https://github.com/DUNE-DAQ/hdf5libs/pull/139
https://github.com/DUNE-DAQ/trigger/pull/437

eflumerf (approved):

wesketchum (approved):
Looks good to me. I think there will be some more significant refactoring work as we get into the details of HDF5libs, as there are more pieces there that could be refactored out to common DataStore or FileDataStore ... but those make sense as part of a future work, with a more solid foundation in place first.

jcfreeman2:
n.b. the single repo CI workflow failed because of a false positive from the linter in which it complained there isn't a copyright notice in `FileDataStoreImpl.hpp` (there is).

PR #498, opened 2026-07-16
Fixed the disabling of the data writing in the TPStreamWriterModule

bieryAtFnal:
# Description

Some time ago, Michal reported that the `--disable-data-storage` option that is part of the run control `start` command was **_not_** causing TPStream file writing to be disabled (even though it disables the writing of raw data files).

It turns out that the "output disable" logic had not been included in the `TPStreamWriterModule`.  This PR fixes that.

In addition, the change in this PR add the checking of the number of TPStream files written in the `disabled_output_test.py` regression test, which was another oversight.  With the changes in this PR, this regression test runs successfully *and* verifies that there are no TPStream files written to disk when the `--disable-data-storage` flag is used.

Here are suggested instructions for testing these changes:

```
DATE_PREFIX=`date '+%d%b'`
TIME_SUFFIX=`date '+%H%M'`

setup_dbt latest
dbt-create -n NFD_DEV_260716_A9 ${DATE_PREFIX}FDDevTest_${TIME_SUFFIX}

git clone https://github.com/DUNE-DAQ/dfmodules.git -b kbiery/tpsw_output_disable

cd ..

. ./env.sh
dbt-build -j 12
dbt-workarea-env

dunedaq_integtest_bundle.sh -r dfmodules -k disable

echo ""
echo -e "\U1F535 \U2705 Note that the disabled_output_test ran successfully, including the correct number of TPStream files. \U2705 \U1F535"
echo ""
echo ""
```

## Type of change

## Testing checklist

## Further checks

I will `clang-format` the TPStreamWriter code as a separate step...

### DIFF
--- a/integtest/disabled_output_test.py
+++ b/integtest/disabled_output_test.py
@@ -203,8 +203,12 @@ def test_data_files(run_dunerc):
     all_ok = True
     # Run some tests on the output data file
     all_ok &= len(run_dunerc.data_files) == expected_number_of_data_files
+    if all_ok:
+        if run_dunerc.verbosity_helper.compare_level(IntegtestVerbosityLevels.drunc_transitions):
+            print(f"\N{WHITE HEAVY CHECK MARK} The correct number of raw data files was found ({expected_number_of_data_files})")
+    else:
+        print(f"\N{POLICE CARS REVOLVING LIGHT} An incorrect number of raw data files was found, expected {expected_number_of_data_files}, found {len(run_dunerc.data_files)} \N{POLICE CARS REVOLVING LIGHT}")
 
-    all_ok = True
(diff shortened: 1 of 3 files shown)


eflumerf (approved):
Looks good. Updated integration test works fine

MRiganSUSX (approved):
@bieryAtFnal thanks for fixing this.

I can confirm that both the integtest and manual test at ehn1 are working as expected. 

The only slightly confusing thing (by name, but not by logic) is this line: https://github.com/DUNE-DAQ/dfmodules/pull/498/changes#diff-e0e9ec26380105e172909297cd78764ed16f891a809a7c78d7c83a0bb56c83aeR231 where it could be confusing to set expectation on TPG based on data files (by name only).

